PR #103, opened 2024-03-13
Feature/redbox 73 unit test core api

gecBurton:
## Context

We want to unit test the core-api so that we have confidence that the code is working as expected.

1. I have added unit test for the core-api
2. I have renamed the core-api endpoints to make the more RESTful https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/
3. In other parts of the code we were (I was!) erroneously force loading the .env.test when we should be leaving it up to `pydantic-settings` to work put where the `.env` file is 

## Guidance to review

- [ ] do the new tests make sense

Note - there is a lot of duplication of code between the various conftests - I have ignored this for now as I am not sure what we should do about it, if anything.

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-73

## Things to check

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - redbox/**
       - embed/**
+      - core_api/**
       - ingest/**
       - Makefile
       - .github/**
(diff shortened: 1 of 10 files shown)


lmwilkigov:
I like the look of the tests! Just did a minor fix in the actions YAML to call them.

lmwilkigov (approved):

PR #106, opened 2024-03-14
REDBOX 76: Adds File status updates to core-api upload

rachaelcodes:
## Context

While there was already an upload method in the `core-api`, there was no way for the front end to find out whether the file had been ingested or embedded. In the future, we anticipate a polling mechanism that can check on the status of a particular file upload. 

* adds a status to the `File` model, which will allow the front end to give information to users about the status of their upload
* updates the docs in line with the new endpoints

## Guidance to review
Do the File statuses make sense? Are the trigger points for updating them in the right place?

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-76 

## Things to check

### DIFF
--- a/core_api/src/app.py
+++ b/core_api/src/app.py
@@ -90,7 +90,11 @@ def health():
     uptime = datetime.now() - start_time
     uptime_seconds = uptime.total_seconds()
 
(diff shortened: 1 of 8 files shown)


gecBurton (approved):
LGTM

:+1: for the verbose tests :)

PR #108, opened 2024-03-14
Feature/redbox 74 trigger embedding post ingest

gecBurton:
## Context

As a User I want the Ingest-app to add messages to the `embed-queue` once it has finished chunking so that the `embed` app knows it should embed the chunks

I have added code to publish ingested chunks to the embed queue

## Guidance to review

Note that the rabbitmq tests are very flakey in so much that there is no fixture to ensure that the messages created in test do not overlap with those in your local state (obv not a problem in CI). this is probably the answer https://pypi.org/project/pytest-rabbitmq/ but is out of scope

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-74

## Things to check

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -163,7 +163,6 @@ services:
       - elasticsearch
     restart: unless-stopped
   rabbitmq:
-    user: "423642704:849146750"
     image: rabbitmq:3-management-alpine
     ports:
       - 5672:5672
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -163,6 +163,7 @@ services:
       - elasticsearch
     restart: unless-stopped
   rabbitmq:
+    user: "423642704:849146750"
     image: rabbitmq:3-management-alpine
     ports:
       - 5672:5672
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -163,6 +163,7 @@ services:
       - elasticsearch
     restart: unless-stopped
   rabbitmq:
+    user: "423642704:849146750"
```
gecBurton, line 166:
```suggestion
```

ingest/src/app.py
```diff
@@ -43,7 +47,17 @@ def ingest_file(self, file: File):
 
         logging.info(f"Writing {len(chunks)} chunks to storage for file uuid: {file.uuid}")
 
-        return self.file_destination.write_items(chunks)
+        items = self.file_destination.write_items(chunks)
+        logging.info(f"written {len(items)} chunks to elasticsearch")
+
+        for chunk in chunks:
```
gecBurton, line 53: this is the key change

PR #109, opened 2024-03-14
Add document upload page

KevinEtchells:
This is adding a basic document upload page, visible at `/upload`. I've made a start on `views.py`. I've also updated `base.html` for the elements shared across pages.

image

## How to test

* Check that the page looks okay
* After uploading, the document name should be displayed in the terminal

## Things not covered
* I'm aware of the missing logo in the footer. This can be a separate job.
* The cancel button/link is currently broken, but I'll be looking at that page next!

## Jira link

https://technologyprogramme.atlassian.net/jira/software/projects/REDBOX/boards/409?selectedIssue=REDBOX-71

### DIFF
--- a/django_app/frontend/style.scss
+++ b/django_app/frontend/style.scss
@@ -1,4 +1,33 @@
 $govuk-assets-path: "govuk-assets/";
 $govuk-font-family: sans-serif;
+$iai-pink: #b62777;
 
-@import "django_app/frontend/node_modules/govuk-frontend/dist/govuk/all";
+@import "../node_modules/govuk-frontend/dist/govuk/all";
+
+
+/* Classification banner */
+.iai-classification-banner {
+  background-color: #2b71c7;
+  color: white;
+  font-size: 0.875rem;
+  margin-bottom: 0.25rem;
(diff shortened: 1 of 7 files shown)


django_app/frontend/style.scss
```diff
@@ -1,4 +1,33 @@
 $govuk-assets-path: "govuk-assets/";
 $govuk-font-family: sans-serif;
+$iai-pink: #b62777;
 
-@import "django_app/frontend/node_modules/govuk-frontend/dist/govuk/all";
+@import "../node_modules/govuk-frontend/dist/govuk/all";
```
rachaelcodes, line 5: I think this is causing tests to fail - the previous version is safer

KevinEtchells: Oh, I didn't mean to change that - I think that's a mistake when resolving conflicts. I'll sort now - thanks!

KevinEtchells: I'm getting this error when changing. Do you get the same?
Error: File to import not found or unreadable: django_app/frontend/node_modules/govuk-frontend/dist/govuk/all

PR #111, opened 2024-03-14
Feature/redbox 44 test end to end file processing from core api to elastic

gecBurton:
## Things to check


docker-compose.yml
```diff
@@ -163,6 +163,7 @@ services:
       - elasticsearch
     restart: unless-stopped
   rabbitmq:
+    user: "423642704:849146750"
```
gecBurton, line 166:
```suggestion
```

PR #119, opened 2024-03-18
Feature/redbox 46 migrate legacy app into streamlit test rig

lmwilkigov:
Have removed some redundant code and renamed the streamlit app


streamlit_app/utils.py
```diff
@@ -159,7 +137,7 @@ def init_session_state() -> dict:
     if "embedding_model" not in st.session_state:
```
gecBurton, line 137: this code is in `model_db.py`? one for another day i think

gecBurton (approved):
LGTM

PR #127, opened 2024-03-19
end-to-end test

gecBurton:
## Context

As an Engineer I want an e2e test to know that a change i have made in one service has not adversely affected another. 

1. A new test has been added that tests that on uploading a file at least one embedding happens
2. `embed_sentences` has been moved in `model_db.py`

Note: Whilst trying to remove all the hardcoded model names I have identified a problem whereby the docker image always preloads the default `paraphrase-albert-small-v2` so any attempt to use a smaller model actually makes this worse by adding another, albeit smaller, model. I have removed the smaller model until we find a way to resolve this.

It is not possible to run this test in CI at this time as the disk runs out of space.

## Guidance to review

```commandLine
docker compose down
cp .env.example .env
docker compose up -d ingest embed core-api
make test-integration
```

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-44

## Things to check


embed/Dockerfile
```diff
@@ -22,5 +22,4 @@ RUN poetry run download-model
 
 EXPOSE 5000
 
-# Run FastAPI
-CMD ["poetry", "run", "uvicorn", "app:app", "--host", "0.0.0.0", "--port", "5000"]
+CMD ["poetry", "run", "python", "app.py"]
```
gecBurton, line 25: this is a big fix. and a good example of why we need an integration test

PR #121, opened 2024-03-18
Feature/docker opt

lmwilkigov:
## Context

Some of the docker images are overly heavy, leading to long builds and lots of RAM consumption

## Changes proposed in this pull request

Original
======
```
redbox-ingest                       latest                d7385c49a8dc   7 minutes ago    3.94GB
redbox-streamlit-app                latest                af474fc7a8f3   11 minutes ago   4.42GB
redbox-embed                        latest                676dfaac7228   3 days ago       3.47GB
redbox-tests                        latest                c7bd1067fc9a   3 days ago       2.08GB
redbox-core-api                     latest                3704ed9f2358   3 days ago       2.1GB
```

CURRENT SIZE TOTAL: 16.01 GB

OPTIMISATION 1: Remove streamlit-app deps from installs (These got left in accidentally)
==============================
```
redbox-ingest                       latest                afb627d3cb28   9 minutes ago       3.62GB (-0.32 GB)
redbox-streamlit-app                latest                98a5cd226a4f   4 minutes ago       4.42GB (0 GB)
redbox-embed                        latest                35b8b6bf549e   6 minutes ago       3.02GB (-0.45 GB)
redbox-tests                        latest                729677e25020   7 minutes ago       1.64GB (-0.44 GB)
redbox-core-api                     latest                80b48e6f5b81   7 minutes ago       1.62GB (-0.48 GB)
```

SAVING: 1.69 GB

OPTIMISATION 2: Build Streamlit app with Buildkit (multistage build process with copied venv, see links)
============================

`redbox-streamlit-app                latest                76e848f9fdcf   8 minutes ago       2.33GB`

SAVING: 2.09 GB

============================
TOTAL OPTIMISATION SAVING = 3.78 GB

## Relevant links
Guide on the optimisations
https://medium.com/@albertazzir/blazing-fast-python-docker-builds-with-poetry-a78a66f5aed0

## Things to check

- [x] I have added any new ENV vars in all deployed environments
- [x] I have tested any code added or changed

PR #133, opened 2024-03-21
isort + ruff

252afh:
## Context

Some more changes to break Liam's PR down slightly

- Some more formatting changes
- Reducing poetry groups used

## Guidance to review

Check that the apps still run

## Relevant links

#121


gecBurton (approved):
looks like this is mainly black. LGTM

PR #134, opened 2024-03-21
e2e test added

gecBurton:
## Context

As an Engineer I want an e2e test to know that a change i have made in one service has not adversely affected another. 

1. A new test has been added that tests that on uploading a file at least one embedding happens
2. `embed_sentences` has been moved in `model_db.py`

Note: Whilst trying to remove all the hardcoded model names I have identified a problem whereby the docker image always preloads the default `paraphrase-albert-small-v2` so any attempt to use a smaller model actually makes this worse by adding another, albeit smaller, model. I have removed the smaller model until we find a way to resolve this.

It is not possible to run this test in CI at this time as the disk runs out of space.

## Guidance to review

```commandLine
docker compose down
cp .env.example .env
docker compose up -d ingest embed core-api
make test-integration
```

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-44

## Things to check


embed/src/app.py
```diff
@@ -48,18 +51,21 @@ def embed_queue_item(self, queue_item: EmbedQueueItem):
         """
 
         chunk: Chunk = self.storage_handler.read_item(queue_item.chunk_uuid, "Chunk")
-        embedded_sentences = embed_sentences(queue_item.model, [chunk.text])
+        embedded_sentences = model_db.embed_sentences(queue_item.model, [chunk.text])
         if len(embedded_sentences.data) != 1:
-            raise ValueError(f"expected just 1 embedding but got {len(embedded_sentences.data)}")
+            logging.error(f"expected just 1 embedding but got {len(embedded_sentences.data)}")
```
252afh, line 56: I'm not too familiar (at all) with how embeddings work. Would having <1 or >1 result here be a critical issue with the program or just a case of "whoopsie that wasn't right"?

embed/src/app.py
```diff
@@ -48,18 +51,21 @@ def embed_queue_item(self, queue_item: EmbedQueueItem):
...
+            return
         chunk.embedding = embedded_sentences.data[0].embedding
-        self.storage_handler.update_item(queue_item.chunk_uuid, chunk)
+        self.storage_handler.update_item(chunk.uuid, chunk)
+        logging.info(f"embedded: {chunk.text}")
```
252afh, line 60: How big could these chunks be, are we going to flood the logs here? Would it be useful to have the uuid of the chunk instead if that's the case?

model_db.py
```diff
@@ -1,10 +1,12 @@
 import collections
 import logging
 import os
+from uuid import uuid4
```
252afh, line 4: Is there a difference between `uuid4` and another `uuid` package we import elsewhere in the program? I think the other one is just `import UUID`?

model_db.py
```diff
@@ -37,3 +39,26 @@ def init_from_disk(self, filepath: str = "models"):
             model = model_name.split("--")[-1]
             self[model] = SentenceTransformer(model_path)
             log.info(f"Loaded model {model}")
+
+    def embed_sentences(self, model: str, sentences: list[str]):
+        model_obj = self[model]
```
252afh, line 44: Do we need a check here for that model existing? And committing to logs if it doesn't

model_db.py
```diff
@@ -37,3 +39,26 @@ def init_from_disk(self, filepath: str = "models"):
...
+            )
+            for i, embedding in enumerate(embeddings)
+        ]
+
+        output = EmbeddingResponse(
+            object="list",
+            data=reformatted_embeddings,
+            embedding_id=str(uuid4()),
+            model=model,
+            model_info=self.get_model_info(model),
```
252afh, line 61: I wonder if we need this `model` property if we're also giving back the`model_info` that includes the model name as well?

252afh: We can leave it in by all means if you think it would be useful by consuming functions, just spitballing

pyproject.toml
```diff
@@ -64,15 +65,21 @@ boto3-stubs = "^1.34.67"
...
-html2markdown = "^0.1.7"
 
 [tool.poetry.group.ai.dependencies]
 anthropic = "^0.21.1"
 litellm = "^1.32.7"
 openai = "^1.14.2"
 
 
+[tool.poetry.group.streamlitapp.dependencies]
+cognitojwt = "^1.4.1"
```
252afh, line 76: Any need for `pandas` or `cognitojwt` in the other poetry group?

tests/test_e2e.py
```diff
@@ -0,0 +1,31 @@
...
+    Given that I have removed all existing chunks from ES
+    When I POST a file to core-api/file
+    I Expect a Chunk with a non-null embedding ... eventually
+    """
+    for chunk in elasticsearch_storage_handler.read_all_items("Chunk"):
+        elasticsearch_storage_handler.delete_item(chunk.uuid, "Chunk")
+
+    file_name = os.path.basename(file_pdf_path)
+    files = {"file": (file_name, open(file_pdf_path, "rb"), "application/pdf")}
+    response = requests.post(url="http://localhost:5002/file", files=files)
```
252afh, line 19: Do we have this url in env vars anywhere? Ignore if not

252afh (approved):

tests/test_e2e.py
gecBurton: good Q.... its specific to the integration test, if the tests were run inside a container then the host would be `elasticsearch` and maybe this should be the case?

core_api/tests/test_app.py
```diff
@@ -91,7 +91,7 @@ def test_ingest_file(app_client, rabbitmq_channel, stored_file, elasticsearch_st
     assert msg["text_hash"] == response.json()["text_hash"]
 
 
-def test_read_all_models(client, example_model_db):
+def test_read_all_models(client):
     """
     Given that I have one model, paraphrase-albert-small-v2, in the database
```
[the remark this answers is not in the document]
gecBurton: i think ill just remove it now that it isnt hard coded

embed/src/app.py
gecBurton: the [SentenceTransformer.encode](https://www.sbert.net/docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode) always returns a list of embeddings, given that we have only entered one item of text i am checking that only one embedding is returned

model_db.py
gecBurton: we have:
* `from uuid import UUID` for the class
* `from uuid import uuid4` for one of the (many) "constructors"

model_db.py
gecBurton: yes we should. ill raise a ticket

model_db.py
gecBurton: @lmwilki will explain! (i think this is a standard way of representing the response)

pyproject.toml
gecBurton: no, this is a merge error. i will remove it

gecBurton: [1d83a24](https://github.com/i-dot-ai/redbox-copilot/pull/134/commits/1d83a24ea5c70d162aaeecb7a8ee49f5e464757f)

embed/src/app.py
gecBurton: they should be small... but you are right, this was just for local debugging tbh. will remove

download_embedder.py
```diff
@@ -4,6 +4,7 @@
 
 from sentence_transformers import SentenceTransformer
 
+
```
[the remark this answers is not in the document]
gecBurton: [b33d5ee](https://github.com/i-dot-ai/redbox-copilot/pull/134/commits/b33d5eec35a1263f7543d498fa38dd58cd2f1140)

Makefile
```diff
@@ -39,6 +39,12 @@ test-ingest:
 	poetry run pytest ingest/tests --cov=ingest -v --cov-report=term-missing --cov-fail-under=40
 
 test-django:
+	docker compose up -d db --wait db
```
gecBurton, line 42:
```suggestion
```

tests/test_e2e.py
gecBurton: worth a ticket to decide how the tests should be run

core_api/tests/test_app.py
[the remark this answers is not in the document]
252afh: :dab:

embed/src/app.py
252afh: ok, in that case I think the way you've done it with a logger instead of causing an exception is the right way to go 👍

model_db.py
252afh: Ah ok, so different use cases, I was worried about slightly different functionality. But then again, there's not much variance a uuid can have, so I doubt it would cause an issue

tests/test_e2e.py
252afh: Agreed 👍

PR #143, opened 2024-03-22
moved to faststream

gecBurton:
## Context

faststream is an abstraction layer for various queuing technologies, by using this instead of `pika` directly we can:
* remove the need to handle connections, event etc
* make use of faststream's in memory queue for testing which makes CI considerably faster as we don't have to install rabbitmq
* more easily move to another queuing technology like Redis if we need to 

* faststream is installed 
* the `ingest` and `embed` workers have been rewritten 
* explicit reference to `pika` has been removed 

Also, the .env was missing from the Dockerfile(s) which caused the e2e tests to fail

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-99

## Things to check

- [x] I have run integration tests locally


lmwilkigov:
Oh gosh, this looks fantastic. So clean ✨

.github/workflows/python-app.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         mkdir -p data/elastic/
         chmod 777 data/elastic/
         cp .env.test .env
-        docker compose up -d --wait elasticsearch minio rabbitmq
+        docker compose up -d --wait elasticsearch minio
```
gecBurton, line 51: no need for this in CI anymore

Makefile
```diff
@@ -45,8 +45,6 @@ test-django:
 test-integration:
 	poetry install --no-root --no-ansi --with dev --without ai,streamlit-app,api,django-app,pytest-django,worker,ingest
 	poetry run pytest tests
-	docker compose up -d --wait db
```
gecBurton, line 48: fixing an unrelated error

embed/src/worker.py
```diff
@@ -0,0 +1,57 @@
...
+async def lifespan(context: ContextRepo):
+    es = env.elasticsearch_client()
+    storage_handler = ElasticsearchStorageHandler(es_client=es, root_index="redbox-data")
+    model_db = SentenceTransformerDB()
+
+    context.set_global("storage_handler", storage_handler)
+    context.set_global("model_db", model_db)
+
+    yield
+    # Clean up the ML models and release the resources
```
gecBurton, line 34:
```suggestion
```

ingest/src/worker.py
```diff
@@ -0,0 +1,86 @@
...
+    storage_handler = ElasticsearchStorageHandler(es_client=es, root_index="redbox-data")
+    model_db = SentenceTransformerDB()
+    chunker = FileChunker(embedding_model=model_db[env.embedding_model])
+
+    context.set_global("s3_client", s3_client)
+    context.set_global("storage_handler", storage_handler)
+    context.set_global("chunker", chunker)
+
+    yield
+    # Clean up the ML models and release the resources
```
gecBurton, line 40:
```suggestion
```

core_api/Dockerfile
```diff
@@ -12,24 +14,18 @@ ENV POETRY_NO_INTERACTION=1 \
 ADD redbox/ /app/redbox
 ADD pyproject.toml poetry.lock /app/
 ADD download_embedder.py /app/
-ADD redbox/ /app/redbox
 
 RUN --mount=type=cache,target=$POETRY_CACHE_DIR poetry install --no-root --no-ansi --with api --without ai,ingest,dev,worker,streamlit-app
 
-FROM python:3.11-slim-buster as runtime
```
gecBurton, line 19: this contradicts everything above it

download_embedder.py
```diff
@@ -1,8 +1,34 @@
+import argparse
```
gecBurton, line 1: largely brought this back exactly as it was

ingest/Dockerfile
```diff
@@ -9,24 +11,18 @@ ENV POETRY_NO_INTERACTION=1 \
 
 WORKDIR /app/
 ADD pyproject.toml poetry.lock ./
-ADD download_embedder.py /app/
 
 RUN --mount=type=cache,target=$POETRY_CACHE_DIR poetry install --no-root --no-ansi --with worker,ingest --without ai,dev,api,streamlit-app
 
-FROM python:3.11-slim-buster as runtime
```
gecBurton, line 16: this contradicts everything above it

redbox/model_db.py
```diff
@@ -5,23 +5,21 @@
...
 
 MODEL_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "models")
 
 log = logging.getLogger()
 log.setLevel(logging.INFO)
 
-env = Settings()
-
 
 class SentenceTransformerDB(collections.UserDict):
```
gecBurton, line 17: i have removed the dependancy on the Settings here so that it can be called without an `.env` existing, e.g. at build time

lmwilkigov (approved):

PR #149, opened 2024-03-26
Add notebook for testing embedding speeds

lmwilkigov:
Adding a notebook that measures some of the speed of embedding models from sentence transformers.


gecBurton (approved):
:+1:

PR #151, opened 2024-03-26
Feature/get file chunks

gecBurton:
## Context

As an User I want a way to retrieve Chunks associated with a File so that I can see that the chunks:
1. are being produced
2. are being embedded

A new endpoint `/file/{file_uuid}/chunks` has been added that returns `list[Chunk]`

## Things to check

- [ ] I have run integration tests locally


redbox/storage/elasticsearch.py
```diff
@@ -122,3 +125,15 @@ def list_all_items(self, model_type: str):
...
         return uuids
+
+    def get_file_chunks(self, parent_file_uuid: UUID) -> list[Chunk]:
+        """get chunks for a given file"""
+        target_index = f"{self.root_index}-chunk"
+
+        res = [
+            item["_source"]
+            for item in self.es_client.search(
+                index=target_index, query={"bool": {"filter": [{"term": {"parent_file_uuid": str(parent_file_uuid)}}]}}
```
lmwilkigov, line 136: We should use the scan functionality shown further up to ensure proper retrieval of large results.

lmwilkigov (changes requested):

redbox/storage/elasticsearch.py
gecBurton: that would explain why my test didn't pass!

lmwilkigov (approved):

lmwilkigov (approved):

PR #157, opened 2024-03-27
REDBOX-71 file upload

rachaelcodes:
## Context

Since starting this ticket, we've discussed moving the S3 upload from the `core-api` to the Django frontend, and instead passing the file by reference.

Links the Django file upload page to the core-api.
Moves the S3 upload and signed URL collection from the core-api to the django application.

## Guidance to review
We're still in the process of updating the architecture, so the key question is: is the separation of concerns sensible?

## Things to check

- [ ] I have run integration tests locally https://github.com/i-dot-ai/redbox-copilot/actions/runs/8571323119


core_api/src/app.py
```diff
@@ -112,38 +111,22 @@ def health():
...
 @app.post("/file", response_model=File, tags=["file"])
-async def create_upload_file(file: UploadFile, ingest: bool = True) -> File:
+async def create_upload_file(name: str, type: str, location: str, ingest=True) -> File:
     """Upload a file to the object store and create a record in the database
 
     Args:
-        file (UploadFile): The file to upload
+        name (str): The file name to be recorded
+        type (str): The file type to be recorded
+        location (str): The presigned file resource location
```
rachaelcodes, line 120: Is there a better data type for the location than 'str'? I couldn't import anything useful from Pydantic

core_api/src/app.py
```diff
@@ -154,6 +137,7 @@ async def create_upload_file(file: UploadFile, ingest: bool = True) -> File:
     if ingest:
         await ingest_file(file_record.uuid)
 
+    # TODO: return something more sensible to the user (including errors)
```
rachaelcodes, line 140: For a 200, what should we return to the user - just 'your file has been uploaded successfully?'

django_app/docker/web/Dockerfile
```diff
@@ -22,6 +22,9 @@ RUN --mount=type=cache,target=$POETRY_CACHE_DIR  poetry install --no-ansi --with
 
 FROM python:3.11-slim-buster as runtime
 
+RUN apt-get update
+RUN apt-get install -y libmagic1
+
```
rachaelcodes, line 27: TODO: remove this (it was needed for a file type checking process that I now think would be better served by a call to an external service)

core_api/src/app.py
gecBurton: https://docs.pydantic.dev/2.0/usage/types/file_types/ ?

core_api/src/app.py
```diff
@@ -112,38 +111,22 @@ def health():
 
 
 @app.post("/file", response_model=File, tags=["file"])
-async def create_upload_file(file: UploadFile, ingest: bool = True) -> File:
+async def create_upload_file(name: str, type: str, location: str, ingest=True) -> File:
```
gecBurton, line 114: note-to-self, this is going to break our integration test

core_api/src/app.py
gecBurton: i dont see why not, even nothing would be fine?

django_app/tests/test_views.py
```diff
@@ -1,8 +1,38 @@
...
 
 
 @pytest.mark.django_db
 def test_declaration_view_get(peter_rabbit, client):
     client.force_login(peter_rabbit)
     response = client.get("/")
     assert response.status_code == 200, response.status_code
+
+
+def count_s3_objects(s3_client):
```
gecBurton, line 12:
```suggestion
def count_s3_objects(s3_client) -> int:
```

gecBurton: `count_s3_objects` is limited to 1,000 objects, could this be a problem?

gecBurton (commented):
have we rules out django-storages[s3]?

django_app/tests/test_views.py
rachaelcodes: Not sure if it would be for testing on GitHub, but I guess it could be running locally. I'll look into the paginated results

rachaelcodes:
> have we ruled out django-storages[s3]?

no, I'll take a look

core_api/src/app.py
```diff
@@ -99,39 +101,25 @@ def health():
     return output
 
 
-@app.post("/file", response_model=File, tags=["file"])
-async def create_upload_file(file: UploadFile, ingest: bool = True) -> File:
+@app.post("/file", response_model=uuid.UUID, tags=["file"])
```
gecBurton, line 104:
```suggestion
@app.post("/file", response_model=uuid.UUID, tags=["file"])
```

core_api/src/app.py
```diff
@@ -99,39 +101,25 @@ def health():
...
+async def create_upload_file(
+    name: str, type: str, location: AnyHttpUrl, ingest=True
+) -> File:
```
gecBurton, line 107:
```suggestion
) -> uuid.UUID:
```

gecBurton (commented):
~is the endpoint you want to modify not `PUT /file/{file_uuid}/ingest` rather than modifying `POST /file`? (sry i only just realised this!)~

actually now i think of it you have no way of knowing what the `uuid` is, ignore me

gecBurton (approved):
LGTM

PR #162, opened 2024-03-28
fix integration tests

gecBurton:
## Context

The integration tests were broken when we moved from `rabbitmq` to `redis`, this is because the `redis` username/password was incorrectly set up. Fixing the `redis` auth will be tackled in another ticket, this PR just gets the tests passing again.

1. `redis` auth has been removed
2. error in integration tests has been corrected
3. some minor corrections to the README has been made 

## Guidance to review

does `make test-integration` work for you?

## Things to check

- [x] I have run integration tests locally

### DIFF
--- a/redbox/models/settings.py
+++ b/redbox/models/settings.py
@@ -42,7 +42,6 @@ class Settings(BaseSettings):
 
     redis_host: str = "redis"
     redis_port: int = 6379
-    redis_password: str = "guest"
 
     dev_mode: bool = False
     django_settings_module: str = "redbox_app.settings"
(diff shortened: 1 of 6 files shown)


redbox/models/settings.py
```diff
@@ -92,16 +91,7 @@ def s3_client(self):
 
         raise NotImplementedError
 
-    def sqs_client(self):
```
gecBurton, line 95: no longer needed

redbox/models/settings.py
```diff
@@ -42,7 +42,6 @@ class Settings(BaseSettings):
 
     redis_host: str = "redis"
     redis_port: int = 6379
-    redis_password: str = "guest"
```
gecBurton, line 45: i have removed this until we work out how to get auth working

rachaelcodes: Is this mentioned somewhere on JIRA as something to fix?

PR #163, opened 2024-03-28
Feature/integration tests in ci using redis

gecBurton:
## Context

We would like integration tests to run in CI.

1. a new workflow has been added in its own .env.integration with a small model to keep the image size down
2. as s3 bucket is created when the s3 client is created (this leads to some simplification in the tests)
3. the test that GETs the chunks has been made more tolerant of failure in case the chunk doesnt exist
4. as this is a long running test it will be run:
    * on a cron schedule for main
    * on demand for your branch

## Guidance to review

- [ ] does CI pass?

Also note that we can delete the `on pull` part of the workflow once this code has been merged into main but not beforehand (if we want to see it running before we merge) this is because this is a new workflow and github wont recognise it otherwise.

## Things to check

- [ ] I have run integration tests locally


.github/workflows/integration.yml
```diff
@@ -0,0 +1,70 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Integration Tests
+
+on:
+  push:
+    branches:
+      - 'feature/**'
```
gecBurton, line 9: this will be deleted post merge

.github/workflows/integration.yml
```diff
@@ -0,0 +1,70 @@
...
+      uses: abatilo/actions-poetry@v2
+      with:
+        poetry-version: 1.7.0
+
+    - name: Build Containers
+      run: |
+        mkdir -p data/elastic/
+        chmod -R 777 data/
+        cp .env.integration .env
+        docker compose up -d --wait core-api embedder ingester miniocreatebuckets
```
gecBurton, line 59:
```suggestion
        docker compose up -d --wait core-api embedder ingester
```

PR #164, opened 2024-03-28
feature/model-improvements

gecBurton:
## Context

We can make more use of pydantic's models:
* support for UUIDs, datetimes, urls, etc
* no need to generate fake values when defaults exist
* can use descriptions to aid understanding

This work is incomplete, but this change is already big enough

many and various

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-94

## Things to check


redbox/models/file.py
```diff
@@ -20,23 +21,45 @@ class ProcessingStatusEnum(str, Enum):
...
+    PPTX = "pptx"
+    TSV = "tsv"
+    XLSX = "xlsx"
+
+
 class File(PersistableModel):
-    path: str
-    type: str
-    name: str
+    path: str = Field(description="location of file")
```
gecBurton, line 49:
```suggestion
    url: str = Field(description="location of file")
```

redbox/models/file.py
```diff
@@ -20,23 +21,45 @@ class ProcessingStatusEnum(str, Enum):
...
+    type: ContentType = Field(description="content_type of file")
```
gecBurton, line 50:
```suggestion
    content_type: ContentType = Field(description="content_type of file")
```

redbox/models/file.py
```diff
@@ -20,23 +21,45 @@ class ProcessingStatusEnum(str, Enum):
...
+    name: str = Field(description="file name")
     storage_kind: str = "local"
-    text: str = ""
-    processing_status: ProcessingStatusEnum = ProcessingStatusEnum.uploaded
+    text: Optional[str] = Field(description="file content", default=None)
```
gecBurton, line 53: we could set this to be a required field if we had an `IngestQueueItem` to reference the file rather than reusing this class?

lmwilkigov (approved):

PR #182, opened 2024-04-04
REDBOX-119 Document Status Endpoint

lmwilkigov:
## Context

We needed a way to introspect where in the process a file has gotten to.

I've added a new status endpoint that surfaces how complete the processing is based on the latest update from workers and the presence of fully embedded chunks in Elasticsearch

## Guidance to review

Upload a file to the API, copy the UUID and then spam the status endpoint with that ID to see it change over time into "complete"

## Things to check


lmwilkigov:
Integration test passed here: https://github.com/i-dot-ai/redbox-copilot/actions/runs/8561702903

redbox/storage/elasticsearch.py
```diff
@@ -141,3 +142,44 @@ def get_file_chunks(self, parent_file_uuid: UUID) -> list[Chunk]:
...
+        file = self.read_item(file_uuid, model_type="File")
+        if file.processing_status == "embedding":
+            chunk_count = self._count_chunks(file_uuid)
+            embedded_chunk_count = self._count_embedded_chunks(file_uuid)
+
+            if chunk_count == embedded_chunk_count:
+                file.processing_status = ProcessingStatusEnum.complete
+                self.update_item(file_uuid, file)
+
+        status_body = {
```
gecBurton, line 181: better to return a `FileStatus` model?

redbox/storage/elasticsearch.py
```diff
@@ -141,3 +142,44 @@ def get_file_chunks(self, parent_file_uuid: UUID) -> list[Chunk]:
...
+
+    def get_file_status(self, file_uuid: UUID):
+        file = self.read_item(file_uuid, model_type="File")
+        if file.processing_status == "embedding":
+            chunk_count = self._count_chunks(file_uuid)
+            embedded_chunk_count = self._count_embedded_chunks(file_uuid)
+
+            if chunk_count == embedded_chunk_count:
+                file.processing_status = ProcessingStatusEnum.complete
+                self.update_item(file_uuid, file)
```
gecBurton, line 179: im not sure about this, because this means that the `File.status` gets updates if , and only if, someone/thing calls `get_file_status`. What if this is never called, and a downstream process relies upon the `File.status` being accurate?

some ideas:
1. we could drop the `File.status` field altogether and could ensure that the only way to to get the status is to call `get_file_status` either directly or via the endpoint
2. we could have the `embedder`/`ingester` update the `File.status` field

either way I dont think that asking for the file status should change it.

core_api/src/app.py
```diff
@@ -47,6 +48,30 @@
 storage_handler = ElasticsearchStorageHandler(es_client=es, root_index="redbox-data")
 
 
+mime_type_to_content_type = {
```
gecBurton, line 51: one to discuss with @rachaelcodes

redbox/storage/elasticsearch.py
gecBurton: another random thought... what if the FileStatus model looked like this:

```python

class ChunkStatus(BaseModel):
    uuid: UUID
    embedded: bool
    
class FileStatus(BaseModel):
    uuid: UUID
    chunks: Optional[list[ChunkStatus]]
```

the rationale being:
1. if the file exists then it is uploaded
2. lets do away with the idea of uploading a file that we will never be chunked/embedded, we have no use case for this and even if we do it doesnt need to be handled by the backend?
3. if `File.chunks` is None then this implies that the file is still chunking
4. the backend should return the maximum information, like how many chunks there are and how many have been embedded but the minimum English. Its up to the various front ends to best interpret this for their users

gecBurton (commented):
some thoughts:

1. i really like the thorough testing!
2. im less sure about the strategy ... some thoughts added
3. please could we embed this into the the e2e/integration tests which has its own way of implying the file status

core_api/src/app.py
rachaelcodes: If we are going down this route, we might need to include multiple options for some content types, as browsers can be inconsistent with MIME types (see https://github.com/getgrav/grav/issues/3422 for a discussion of this from PHP-land). Though apparently it's not unheard of for .eml files to be assigned 'application/octet-stream' https://stackoverflow.com/questions/65698647/what-is-the-mime-type-of-eml-files-in-emails which really wouldn't be helpful at all. 

At the very least we should probably add: "text/xml": "xml" https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types . Alternatively, we could take the file extension instead (which is what I chose to do for REDBOX-71)

Given that we should be double-checking user submitted MIME types etc anyway, I think in the future we should move towards a system where we get the MIME types assessed by a service like ClamAV, which should hopefully give us a more standard list (and maybe even the file type).

rachaelcodes: The other thing about REDBOX-71: we're also looking to pass the file type directly to core-api, so this file check wouldn't be happening in core-api any more. https://github.com/i-dot-ai/redbox-copilot/pull/157/files#diff-ba6d55b32fa5e31957b2d97cbe5c55d37989f3b72bd66db3653ef54d9db3f52fR106

gecBurton: what process uses this field in a meaningful way? could we either:
* make it a plain `str`
* drop it all together

redbox/storage/elasticsearch.py
lmwilkigov: Great shout

redbox/storage/elasticsearch.py
lmwilkigov: I like this mental model, much cleaner. Lets have a crack at it on Tuesday

core_api/src/app.py
```diff
@@ -109,12 +134,18 @@ async def create_upload_file(file: UploadFile, ingest: bool = True) -> File:
     Returns:
         File: The file record
     """
+    try:
+        content_type = mime_type_to_content_type[file.content_type]
```
252afh, line 138: Naming is a bit confusing here, you're using the content_type from the file to get the content_type from the mime_type in the dict?

core_api/src/app.py
```diff
@@ -199,6 +230,22 @@ def get_file_chunks(file_uuid: UUID) -> list[Chunk]:
...
+    """Get the status of a file
+
+    Args:
+        file_uuid (str): The UUID of the file to get the status of
+
+    Returns:
+        File: The file with the updated status
+    """
+
+    status = storage_handler.get_file_status(file_uuid)
```
252afh, line 244: Can we add an `unknown` status here for files that don't exist to obfuscate that fact

ingester/src/worker.py
```diff
@@ -77,6 +77,10 @@ async def ingest(
         queue_item = EmbedQueueItem(chunk_uuid=chunk.uuid)
         logging.info(f"Writing chunk to storage for chunk uuid: {chunk.uuid}")
         await publisher.publish(queue_item)
+
+    file.processing_status = ProcessingStatusEnum.embedding
+    storage_handler.update_item(file.uuid, file)
```
252afh, line 82: All of the models given to this function have an id, could we remove it as a separate argument?

gecBurton: this actually is the case in `main`

gecBurton (approved):
nice :)

core_api/src/app.py
```diff
@@ -190,6 +187,24 @@ def get_file_chunks(file_uuid: UUID) -> list[Chunk]:
...
 
+@app.get("/file/{file_uuid}/status", tags=["file"])
+def get_file_status(file_uuid: UUID) -> FileStatus:
+    """Get the status of a file
+
+    Args:
+        file_uuid (str): The UUID of the file to get the status of
+
+    Returns:
+        File: The file with the updated status
```
252afh, line 198: Return type in the comment is wrong

252afh (approved):

ingester/src/worker.py
252afh: thanks

PR #197, opened 2024-04-09
feature/simplify-ingester

gecBurton:
## Context

We can simplify the ingester.

* I have made better use of the `File` object so that we do not need to pass around so many individual variables
* this means that we no longer need to get the same data (twice) from s3 in the ingester

## Guidance to review

- [ ] Does CI pass, including the integration tests?

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/8615444109

### DIFF
--- a/core_api/src/app.py
+++ b/core_api/src/app.py
@@ -114,17 +114,19 @@ async def create_upload_file(name: str, type: str, location: AnyHttpUrl) -> uuid
         UUID: The file uuid from the elastic database
     """
 
-    file_record = File(
+    file = File(
         name=name,
         url=str(location),  # avoids JSON serialisation error
         content_type=type,
     )
 
-    storage_handler.write_item(file_record)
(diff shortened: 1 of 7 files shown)


core_api/src/app.py
```diff
@@ -102,10 +101,8 @@ def health():
     return output
 
 
-@app.post("/file", response_model=uuid.UUID, tags=["file"])
-async def create_upload_file(
-    name: str, type: str, location: AnyHttpUrl, ingest=True
-) -> uuid.UUID:
+@app.post("/file", tags=["file"])
+async def create_upload_file(name: str, type: str, location: AnyHttpUrl) -> uuid.UUID:
```
gecBurton, line 105: i removed `ingest=true` whilst here

lmwilkigov (approved):

PR #198, opened 2024-04-09
add: ecr and make commands

ghannay10:
## Things to check


gecBurton (commented):
all looks good except `AWS_ACCOUNT_ID` being stored in the .env. The envvars are handled by https://docs.pydantic.dev/latest/concepts/pydantic_settings/ so if we include the `AWS_ACCOUNT_ID` here then this field will need to be added to `redbox/models.settings.py`, are we sure we want to do this? could we inject this no-runtime variable another way, if only via another infra specific env file?

PR #201, opened 2024-04-10
feature/file-as-a-reference

gecBurton:
# Do Not Merge - just for research

## Context

Currently the `File` object stored in Elastic is confusing as it stored a lot of data, like the underlying text that we dont actually use. In reality we only actually want to persist a reference to the file in s3. We cannot us the presigned-url for this as it is transient, so instead we store the `key` (the `bucket is an env var) and generate the url on demand.

The `File` is now just the s3 key

## Guidance to review

- [ ] does this fit the schema as set out in https://github.com/i-dot-ai/redbox-copilot/pull/202

## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/8630071296


core_api/src/app.py
```diff
@@ -100,36 +100,40 @@ def health():
 
     return output
 
+class FileRequest(BaseModel):
+    url: AnyHttpUrl
 
 @app.post("/file", tags=["file"])
-async def create_upload_file(name: str, type: str, location: AnyHttpUrl) -> uuid.UUID:
+async def create_upload_file(file: FileRequest) -> File:
```
lmwilkigov, line 107: Should we rename this func to `add_file` or something now that we pass by reference?

lmwilkigov (approved):

core_api/src/app.py
gecBurton: good idea

gecBurton:
this change will break streamlit, so i am holding off merging it until streamlit is in its own repo

gecBurton:
trying another approach

PR #202, opened 2024-04-10
Proposed schema and sequence diagrams

rachaelcodes:
## Context
There's been a number of discussions about where data is stored (especially File metadata), and I realised that members of the team had different ideas about where the discussion ended up.

Additionally, while I was working on the /chat APIs I realised I wasn't fully sure about the data flows.

I've opened up this pull request to allow us to collect data flow and schema discussions and decisions in one place.

I've created diagrams for our APIs and diagrams of schemas for our data stores and stored them .md file in /docs . 

Here are the diagrams:

File /upload data flow
Screenshot 2024-04-10 at 09 08 06

/chat/vanilla data flow
Screenshot 2024-04-10 at 09 08 12

/chat/rag data flow
Screenshot 2024-04-11 at 09 55 34

Django schema
Screenshot 2024-04-11 at 09 55 48

Elastic schema (file and chunks only so far)
Screenshot 2024-04-10 at 09 08 27

## Guidance to review
Have a look at the diagrams, make corrections and suggestions and ask questions. If possible, can you start threads from code comments rather than general PR comments so it's easier to easier to follow threads of discussion. 

Where it turns out we've got massive differences of opinion, then I think it might be best to come to a decision on a team call.

## Things to check


notebooks/sequence_and_schema.ipynb
```diff
@@ -0,0 +1,172 @@
...
+    "    }\n",
+    "    \n",
+    "    FileRecord }|--|| UserGroup: \"FileRecord.owner\"\n",
+    "    FileRecord {\n",
+    "        UUID uuid\n",
+    "        UUID owner\n",
+    "        string key \n",
+    "    }\n",
+    "\n",
+    "    UserGroup ||--|{ Message : \"Mesage.owner\"\n",
```
rachaelcodes, line 100: Do we need Messages to have a concept of an owner, if they are part of a Chat History which also has that owner?

notebooks/sequence_and_schema.ipynb
```diff
@@ -0,0 +1,172 @@
...
+    "    File ||--o{ Chunk : \"File.uuid\"\n",
+    "    File {\n",
+    "        UUID uuid\n",
+    "    }\n",
+    "    Chunk {\n",
+    "        UUID uuid\n",
+    "        UUID parent_file_uuid\n",
+    "        int index\n",
+    "        str text\n",
+    "        dict metadata\n",
```
rachaelcodes, line 144: What's in this metadata?

gecBurton:
what does "Embedding API" mean in `/chat/rag data flow`?

notebooks/sequence_and_schema.ipynb
```diff
@@ -0,0 +1,172 @@
...
+   "metadata": {},
+   "source": [
+    "```mermaid\n",
+    "---\n",
+    "title: Data sequence - POST /chat/rag\n",
+    "---\n",
+    "\n",
+    "sequenceDiagram\n",
+    "    Django->> Core: ChatHistory.messages[], File[].uuid\n",
+    "    Core->>Embedding API: ChatHistory.messages[]\n",
```
gecBurton, line 62: what is the "Embedding API" here? i.e. is it a service?

notebooks/sequence_and_schema.ipynb
gecBurton: it seems reasonable that a message would belong to chat history?

gecBurton: it seems reasonable that a message would belong to chat history?

maybe we could call "Message" -> "ChatMessage"?

notebooks/sequence_and_schema.ipynb
```diff
@@ -0,0 +1,172 @@
...
+    "    }\n",
+    "\n",
+    "    Message }|--|| ChatHistory: \"ChatHistory.messages\"\n",
+    "    UserGroup ||--|{ ChatHistory: \"ChatHistory.owner\"\n",
+    "\n",
+    "    ChatHistory {\n",
+    "        UUID uuid\n",
+    "        string name\n",
+    "        UUID owner\n",
+    "        UUID[] messages\n",
```
gecBurton, line 115: given that these relationships are 1:* should we move the reference to the child class, i.e. Message & File have references to ChatHistory?

notebooks/sequence_and_schema.ipynb
gecBurton: https://github.com/Unstructured-IO/unstructured/blob/4656b8cbe5f1f0c8468a7f11ab8bb39e07fd6c94/unstructured/documents/elements.py#L154

notebooks/sequence_and_schema.ipynb
lmwilkigov: The embedding of user queries (i.e. the hot path as opposed to the async cold path used for chunks) happens inside of Core API. May be if we rename it `Core - Embedding`?

notebooks/sequence_and_schema.ipynb
lmwilkigov: Agree on the `ChatMessage` notation to avoid conflation with Alert messages that we might implement in future. A ChatMessage will always belong to a `ChatHistory`. However, it might make usage metrics easier for individual messages to have ownership by user directly in object. Equally, having individual ownership of `ChatMessages` enables us to explore group chats in future if we need it

notebooks/sequence_and_schema.ipynb
gecBurton: this makes sense 👍

notebooks/sequence_and_schema.ipynb
lmwilkigov: lets rename `files_received` to `files_retrieved` (to match up with RAG wording).

George, could you explain that referencing? Not sure I grasp the concept

gecBurton: > ...explain that referencing...

its better/easier to store 1 parent-reference than many child-references?

lmwilkigov: Ah, yes one parent reference. That'll match up with our File Chunk style relationship

notebooks/sequence_and_schema.ipynb
rachaelcodes: maybe we take out this line and the subsequent one as it's all happening inside Core?

notebooks/sequence_and_schema.ipynb
rachaelcodes: Cool. Would we be directly referencing anything from this metadata in code anywhere?

notebooks/sequence_and_schema.ipynb
lmwilkigov: That works

notebooks/sequence_and_schema.ipynb
gecBurton: agree with both points. lets avoid polymorphism in the database at all costs. for UserGroups lets subclass django's Group model (which i think has a many-to-many relationship to the User)?

notebooks/sequence_and_schema.ipynb
lmwilkigov: Page_number will be used for citation to load up the specific page of a URL. For non paged files (like txt), we're likely to add an offset attribute to chunks (e.g. chunk 22/33 of FileX starts at 3456543 char offset into FileX). That will help us highlight text.

gecBurton: we are in sterreamlit, would be good to get this nailed down, if only the mandatory fields (can always have `extra=true` to allow people to crowbar in additional stuff

notebooks/sequence_and_schema.ipynb
rachaelcodes: Yes to messages, not to files (which could be used in several chats) - but that's my mistake of labelling

gecBurton (approved):

lmwilkigov (approved):

PR #222, opened 2024-04-12
Bugfix/ruff not checking code

gecBurton:
## Context

Our CI check for ruff formatting is not catching errors

1. added `--check` to the `ruff format` command so that it fails rather than trying to fix the code (withour committing the code)
2. `make format` has been run locally

## Guidance to review

This is the only real change https://github.com/i-dot-ai/redbox-copilot/pull/222/files#r1562178510 everything else is a consequence of running `ruff`.

## Things to check


.github/workflows/python-app.yml
```diff
@@ -100,7 +100,8 @@ jobs:
 
     - name: Format code
       run: |
-        make format
+        poetry run ruff format . --check
```
gecBurton, line 103: this is the key change

lmwilkigov (approved):

PR #225, opened 2024-04-13
Feature/redbox 140 finalize file model

gecBurton:
## Context

I have change the `File` model in line with https://github.com/i-dot-ai/redbox-copilot/pull/202

* all attributes of the `File` have been removed apart from the `name` which has been relabelled `key` (it was always the s3-key in reality)
* this has led to a lot of simplification re: the use of the pre-signed url

## Guidance to review

Notice that although not in the spec I have add the `bucket` because:
  * it is the combination of `bucket/key` uniquely (and globally) identifies the file
  * it was easy to do
however, this may not be helpful given that the bucket is also an env-var so am happy to remove this?

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-140

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/8675170952
- [x] I have manually tested the streamlit app

### DIFF
--- a/tests/test_e2e.py
+++ b/tests/test_e2e.py
@@ -23,18 +23,11 @@ def test_upload_to_elastic(file_pdf_path, s3_client):
             ExtraArgs={"Tagging": "file_type=pdf"},
         )
 
-        authenticated_s3_url = s3_client.generate_presigned_url(
-            "get_object",
-            Params={"Bucket": bucket_name, "Key": file_key},
-            ExpiresIn=3600,
-        )
-
         response = requests.post(
             url="http://localhost:5002/file",
             params={
-                "name": "filename",
-                "type": ".pdf",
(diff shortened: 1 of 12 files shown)


tests/test_e2e.py
```diff
@@ -23,18 +23,11 @@ def test_upload_to_elastic(file_pdf_path, s3_client):
...
-        )
-
         response = requests.post(
             url="http://localhost:5002/file",
-            params={
-                "name": "filename",
-                "type": ".pdf",
-                "location": authenticated_s3_url,
+            json={
+                "name": file_key,
```
gecBurton, line 29:
```suggestion
                "key": file_key,
```

lmwilkigov (approved):

PR #227, opened 2024-04-15
Fix browser console errors

KevinEtchells:
## Context

While working on something else, I noticed a couple of browser console errors:

* The gov.uk inline script is being blocked by our Content Security Policy (meaning no gov.uk javascript is running)
* Missing manifest.json file

This update addresses both of these issues.

* Update the CSP to allow the inline script (but only this inline script) by using a hash
* Add the manifest.json file to the project. This alone didn't fully fix the problem, but removing the whitespace from the URL did

## Guidance to review

* Visit any page and check there are no browser console errors
* Please look at the CSP changes in `settings.py` to confirm they are sensible


gecBurton (approved):
good spot!

PR #237, opened 2024-04-16
Added models and migrations based on schema diagram

252afh:
## Things to check


django_app/redbox_app/redbox_core/migrations/0003_chathistory_file_chatmessage_and_more.py
```diff
@@ -0,0 +1,73 @@
...
+        migrations.CreateModel(
+            name='File',
+            fields=[
+                ('id', models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
+                ('created_at', models.DateTimeField(auto_now_add=True)),
+                ('modified_at', models.DateTimeField(auto_now=True)),
+                ('name', models.TextField(max_length=2048)),
+                ('processing_status', models.CharField(choices=[('uploaded', 'Uploaded'), ('parsing', 'Parsing'), ('chunking', 'Chunking'), ('embedding', 'Embedding'), ('indexing', 'Indexing'), ('complete', 'Complete')])),
+                ('key', models.CharField(max_length=2048)),
+                ('file_type', models.CharField(choices=[('.eml', 'Eml'), ('.html', 'Html'), ('.json', 'Json'), ('.md', 'Md'), ('.msg', 'Msg'), ('.rst', 'Rst'), ('.rtf', 'Rtf'), ('.txt', 'Txt'), ('.xml', 'Xml'), ('.csv', 'Csv'), ('.doc', 'Doc'), ('.docx', 'Docx'), ('.epub', 'Epub'), ('.odt', 'Odt'), ('.pdf', 'Pdf'), ('.ppt', 'Ppt'), ('.pptx', 'Pptx'), ('.tsv', 'Tsv'), ('.xlsx', 'Xlsx'), ('.htm', 'Htm')])),
```
lmwilkigov, line 37: The capitalisation looks a bit funky here. Shall we go all lower or UPPER caps?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -42,16 +43,63 @@ class ProcessingStatusEnum(models.TextChoices):
...
+    original_file = models.FileField(storage=settings.BUCKET_NAME)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    @property
+    def file_type(self):
+        name = self.original_file.name
+        return name.split(".")[-1]
+
+    @property
+    def file_url(self):
```
gecBurton, line 80: this is cute. TIL

django_app/redbox_app/redbox_core/models.py
```diff
@@ -42,16 +43,63 @@ class ProcessingStatusEnum(models.TextChoices):
...
         stage = processing_status_list.index(self.processing_status)
         if stage == len(processing_status_list) - 1:
             return self.processing_status
         return f"{stage + 1}/{len(processing_status_list) - 1} {self.processing_status}"
+
+
+class ChatHistory(UUIDPrimaryKeyBase, TimeStampedModel):
+    name = models.TextField(max_length=1024, null=False, blank=False)
+    users = models.ManyToManyField(User)
+    files_received = models.ForeignKey(File, on_delete=models.CASCADE, related_name="files_received")
```
gecBurton, line 98: this is a tough one.... i feel that the relationship between a chat and a file is many-to-many... and perhaps more complex than that... maybe need a through model to label the collection of files, e.g. 'environment" may group files 1, 2, & 3 and "industry" might group file 3,4,5... just a thought, open to discussion

django_app/redbox_app/redbox_core/models.py
```diff
@@ -42,16 +43,63 @@ class ProcessingStatusEnum(models.TextChoices):
     complete = "complete"
 
 
-# TO DO: Based on /redbox/models/file.py, but not complete
-class File(models.Model):
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    path = models.TextField(help_text="location of file")
-    name = models.TextField()
-    processing_status = models.CharField(choices=ProcessingStatusEnum.choices, default=ProcessingStatusEnum.uploaded)
+class FileTypeEnum(models.TextChoices):
```
gecBurton, line 46: do we need this any more?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -42,16 +43,63 @@ class ProcessingStatusEnum(models.TextChoices):
...
+    name = models.TextField(max_length=1024, null=False, blank=False)
+    users = models.ManyToManyField(User)
+    files_received = models.ForeignKey(File, on_delete=models.CASCADE, related_name="files_received")
+    files_retrieved = models.ForeignKey(File, on_delete=models.CASCADE, related_name="files_retrieved")
+
+
+class ChatMessage(UUIDPrimaryKeyBase, TimeStampedModel):
+    chat_history = models.ForeignKey(ChatHistory, on_delete=models.CASCADE)
+    text = models.TextField(max_length=32768, null=False, blank=False)
+    role = models.TextField(max_length=1024, null=False, blank=False)
```
gecBurton, line 105: should this be restricted to some enum?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -42,16 +43,63 @@ class ProcessingStatusEnum(models.TextChoices):
...
+
+    @property
+    def file_url(self):
+        return self.original_file.url
+
+    @property
+    def name(self):
+        return self.original_file.name
 
     def get_processing_text(self) -> str:
```
gecBurton, line 87: i wonder if this needs to be stored at all, I think it is transient?

django_app/redbox_app/redbox_core/models.py
252afh: What does files_received in a chat even mean in the first place? Do users add files to certain chats? I thought it would be more of a case of files_selected for chats, and files_referenced for a message?

django_app/redbox_app/redbox_core/models.py
252afh: Nope, good shout

django_app/redbox_app/redbox_core/models.py
252afh: What is `role` in this context?

django_app/redbox_app/redbox_core/models.py
252afh: I wonder what transactions happen when querying the file for information, whether a request for the file is made, or if the data like file name is stored in the db?

django_app/redbox_app/redbox_core/models.py
gecBurton: i dont know tbh. @lmwilkigov do you know?

django_app/redbox_app/redbox_core/models.py
gecBurton: I think this is https://github.com/i-dot-ai/redbox-copilot/blob/4dc8ebf96b633d2e0c2ea2a4f0387dc9c39361d9/core_api/src/routes/chat.py#L39

django_app/redbox_app/redbox_core/models.py
gecBurton: in the backend this is handled by https://github.com/i-dot-ai/redbox-copilot/blob/4dc8ebf96b633d2e0c2ea2a4f0387dc9c39361d9/redbox/storage/elasticsearch.py#L193

django_app/redbox_app/redbox_core/models.py
rachaelcodes: > I thought it would be more of a case of files_selected for chats, and files_referenced for a message?
This sounds broadly right. If this is the case, I imagine we don't need a list of files_referenced for a whole history that combines the message files_referenced.

* @lmwilkigov I think this also leads to a question about whether we have a blanket search on all files a user has uploaded, or they have the option to choose specific ones?

django_app/redbox_app/redbox_core/models.py
rachaelcodes: It's something expected by various ChatParsers

rachaelcodes (commented):
When discussions around the schema have been resolved, could you also update `docs/architecture/transactions_and_schema.md`? Thanks

django_app/redbox_app/redbox_core/models.py
252afh: George is your feeling for this that it should come from elsewhere in the code? A function that returns the status of the file from checking the file is in s3 and from checking the state of embed/ingest in core?

django_app/redbox_app/redbox_core/models.py
252afh: Are we happy with the changes I've made regarding this?

KevinEtchells: > It looks like all files from recent meetings, but then you'd still need to keep track of the files for a specific chat in case the user uploads more files after-the-fact?

I'm not certain we would? files_selected would future proof us though.

django_app/redbox_app/redbox_core/models.py
KevinEtchells: We'd want to display the original filename on the documents page wouldn't we? Sounds like we'd need it, or are we saying we'd get this from elsewhere?

django_app/redbox_app/redbox_core/models.py
252afh: I think it depends on if we're going to limit the RAG search to the files present in the chat.

I have a chat that has files a, b, c.

I chat with that for a while.

I then go upload files d and e.

I make a new chat and chat there for a bit.

If we have chat history, and I can go back and view the first chat, without saving the files used for that chat there's no way to know what files that chat has access to.

But if we don't limit the RAG to those files anyway, then it doesn't really make a difference because as soon as I enter a new message into chat 1 it now also includes d and e in all new retrievals

django_app/redbox_app/redbox_core/models.py
252afh: We can have the file name, I think George was highlighting the part of the model that got the file processing status. I feel like this is a bit beyond a model function now.

I think it would be the frontend checking the status with s3, and then the frontend checking with core the status of the file in elastic

django_app/redbox_app/redbox_core/models.py
KevinEtchells: I don't think we limit to those files for MVP, so it will include d and e. But it sounds like we will want to in the future. So it depends whether it's easiest to add it in now or wait until later.

django_app/redbox_app/redbox_core/models.py
KevinEtchells: Oh, with you now. Could we have a `processed` boolean, so we're only checking statuses for files that are still processing?

lmwilkigov (approved):

PR #240, opened 2024-04-16
added model to docker build

lmwilkigov:
## Context

Ingester was downloading an OCR model live

Now we do it in build

## Things to check

### DIFF
--- a/ingester/download_ocr_models.py
+++ b/ingester/download_ocr_models.py
@@ -1,0 +1,2 @@
+from unstructured_inference.models.tables import load_agent
+load_agent()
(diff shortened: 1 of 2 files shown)


ingester/download_ocr_models.py
```diff
@@ -0,0 +1,3 @@
+from unstructured_inference.models.tables import load_agent
+
+load_agent()
```
gecBurton, line 3: unlikely to be a problem i know, but a good idea to use the usual:

```python
if __name__ == "__main__":
    load_agent()
```

just in case anyway tries to import this file

gecBurton (approved):
LGTM

ingester/download_ocr_models.py
lmwilkigov: Added, great shout!

PR #242, opened 2024-04-17
feature/rationalise-file-endpoint-post

gecBurton:
## Context

As an integrator I want it to be clear what payload i should be posting to 

before this change it was possible to POST details like the `uuid` which we want to be worked out by the server itself and the swagger doc example payload looked like this
image

now it looks like

image

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/8734771616

### DIFF
--- a/core_api/src/routes/file.py
+++ b/core_api/src/routes/file.py
@@ -3,6 +3,7 @@ from uuid import UUID
 
 from fastapi import FastAPI, HTTPException
 from faststream.redis.fastapi import RedisRouter
+from pydantic import BaseModel, Field
 
 from core_api.src.publisher_handler import FilePublisher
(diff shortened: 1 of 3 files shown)


core_api/src/routes/file.py
```diff
@@ -47,17 +48,26 @@
 file_app.include_router(router)
 
 
+class FileRequest(BaseModel):
+    """Reference to file stored on s3"""
+
+    key: str = Field(description="file key", examples=["sales.csv", "README.txt"])
```
gecBurton, line 54:
```suggestion
    key: str = Field(description="file key", examples=["policies.pdf"])
```

lmwilkigov (approved):

PR #243, opened 2024-04-17
Feature/health endpoints for workers

gecBurton:
## Things to check


ingester/Dockerfile
```diff
@@ -11,7 +11,7 @@ WORKDIR /app/
 ADD pyproject.toml poetry.lock ./
 ADD download_embedder.py /app/
 
-RUN --mount=type=cache,target=$POETRY_CACHE_DIR poetry install --no-root --no-ansi --with worker,ingester --without ai,dev,api
+RUN --mount=type=cache,target=$POETRY_CACHE_DIR poetry install --no-root --no-ansi --with ingester,api --without ai,dev
```
252afh, line 14: `api` seems like a big dependency to have for the ingester, are there some packages we could put in the ingester block instead?

embedder/tests/test_worker.py
```diff
@@ -17,8 +17,18 @@ async def test_embed_item_callback(elasticsearch_storage_handler, embed_queue_it
     unembedded_chunk = elasticsearch_storage_handler.read_item(embed_queue_item.chunk_uuid, "Chunk")
     assert unembedded_chunk.embedding is None
 
-    async with TestRedisBroker(broker) as br, TestApp(app):
+    async with TestRedisBroker(router.broker) as br:
```
252afh, line 20: Is there a `TestRedisRouter` we could use to keep in step with how the embedder works?

gecBurton: unfortunately not :(

ingester/Dockerfile
gecBurton: unfortunately ECS requires us to have an http endpoint, so we need all 3 dependancies in the api group in `core-api`, `embedder` and `ingester`

embedder/tests/test_worker.py
252afh: Ah thanks for the docs :)

252afh (approved):

PR #247, opened 2024-04-19
Build file-status update JavaScript

KevinEtchells:
## Context

This PR covers:
* Adding `Alpine.js` to the project
* Add JavaScript so file statuses can be updated without requiring a page reload

Jira ticket: [REDBOX-139](https://technologyprogramme.atlassian.net/jira/software/projects/REDBOX/boards/409?selectedIssue=REDBOX-139)

### Alpine.js
* We're using the CSP-safe version of Alpine so we don't need to allow `unsafe-eval`. I have added a comment to `alpine.min.js` to make this clear.
* In order to work nicely with `django-compressor`, this isn't using JS modules. I've added a comment to `alpine.min.js` for what to do if we ever want to use this as a module.
* The top lines of `documents.js` is a bit of a workaround to give us the nice typing without using TypeScript and without using JS modules. 

### File status updates
* The endpoint will 404 at the moment. There are 2 constants in `documents.js` that can be updated once we have the file-status endpoint in Django.
* Update to `documents.html` to pass in the document ID to an attribute and initiate Alpine (only if the file processing isn't already complete).
* This work can then be picked up again as part of [REDBOX-77](https://technologyprogramme.atlassian.net/jira/software/projects/REDBOX/boards/409?selectedIssue=REDBOX-77)

## Guidance to review
* As the endpoint isn't running, please just have a look through the code to check this approach is sensible
* You could temporarily remove the `if (response.ok)` check in `documents.js` to see text coming in to the page (albeit a 404 error!)

[REDBOX-139]: https://technologyprogramme.atlassian.net/browse/REDBOX-139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[REDBOX-77]: https://technologyprogramme.atlassian.net/browse/REDBOX-77?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gecBurton (approved):
Looks sensible to me

PR #276, opened 2024-04-25
Fixed chunking so index is set correctly

### LINKED ISSUE #272: `chunk_clustering` sets all indices to 1
**Describe the bug**

All chunks are currently ingested with an index of 1. This means we can't reassemble file text from its constituent chunks. @gecBurton says this is a bug in `chunk_clustering`, specifically [here](https://github.com/i-dot-ai/redbox-copilot/blob/b4b076c6a883e240770d46fc232aac8b1e87e026/redbox/parsing/chunk_clustering.py#L72).

**To Reproduce**

Ingest some data.

```
es.search(
    index="redbox-data-chunk",
    query={
        "match": {
            "parent_file_uuid": {
                "query": "..." # some UUID
            }
        }
    }
)['hits']['hits']
```

**Expected behavior**

Ordered chunk indices.

ghost:
## Context

Fixes #272 . Allows clustered chunks to have the correct index.

Tested in streamlit -- not done any here.

### DIFF
--- a/redbox/parsing/chunk_clustering.py
+++ b/redbox/parsing/chunk_clustering.py
@@ -70,7 +70,7 @@ def cluster_chunks(
                 # if there are multiple chunks in the cluster, merge them
                 new_chunk = Chunk(
                     parent_file_uuid=chunks_in[0].parent_file_uuid,
-                    index=1,
+                    index=i,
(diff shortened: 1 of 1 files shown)


gecBurton (approved):
❤️ 
TVM

PR #297, opened 2024-05-01
Bugfix/fix migration being arranged incorrectly

252afh:
## Context

The migration file that was generated as part of the model's PR was causing issues by creating tables in a weird order.

Re-arranged the migration so relations aren't made after the tables are made

Added minio to required services for django-app

## Guidance to review

`make reset-db` and `docker compose up django-app`, check that migrations apply without error

## Things to check

_It looks like more changes than there are in the migration file because ruff ran automatically on it_


KevinEtchells (approved):
Looks good. And thank you for adding minio in!

PR #301, opened 2024-05-02
feature/REDBOX-170-add-jwt-to-core-api

gecBurton:
## Context

As a backend I want to know who is requesting what so that I can GET/CREATE the right data for them

Note: even though we are using a JWT to encode the user_uuid this is Not used for Authentication, although it maybe used for this in the future with an appropriate code change (i.e. using cognito)

* core-api: All (except /health) endpoints now require a JWT in the headers. The JWT encodes a user_uuid, which is a UUID.
* django: all requests to core-api now send a Bearer token in the header

## Guidance to review

Note that nothing is actually done with the user_uuid in this PR, this will be tackled in another ticket.

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-170

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/8925978665/job/24515954483

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -65,6 +65,7 @@ jobs:
         pip install requests
         pip install pytest
         pip install boto3
+        pip install python-jose
         python -m pytest tests
 
     - name: notify slack failure
--- a/README.md
+++ b/README.md
@@ -83,6 +83,30 @@ This project is licensed under the MIT License - see the [LICENSE](./LICENSE) fi
 
 # Security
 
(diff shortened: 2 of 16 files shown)


lmwilkigov (dismissed):
LGTM on the substance of the code, however, could you add guidance on how a developer can get a valid token so that they might test out the CoreAPI via the webpage

252afh (approved):
Love it

KevinEtchells (approved):

PR #308, opened 2024-05-03
Feature/doc management page

252afh:
## Context

We need to be able to add, view and delete documents that have been saved. This is being done on a dedicated page using a FileField django model field.

- Enabled functionality on the `/documents/` page

## Guidance to review

- Run the app locally, try to add a file, check that it is shown on the view page
- Attempt to remove a document

## Relevant links

[https://technologyprogramme.atlassian.net/jira/software/projects/REDBOX/boards/409?assignee=712020%3Aef73792a-97e3-43d0-9992-3e9d55feb555&selectedIssue=REDBOX-77](https://technologyprogramme.atlassian.net/jira/software/projects/REDBOX/boards/409?assignee=712020%3Aef73792a-97e3-43d0-9992-3e9d55feb555&selectedIssue=REDBOX-77)

## Things to check


252afh:
This works locally, but not yet on dev due to duplicate s3 code and difficulty testing. This should probably be its own PR now though and the dev work can be done in another one as it's a fair sized rework - it might make this PR quite large

django_app/redbox_app/redbox_core/models.py
```diff
@@ -55,22 +55,37 @@ class ProcessingStatusEnum(models.TextChoices):
...
     def name(self):
-        return self.original_file.name
+        return (
+            self.original_file_name
+            if self.original_file_name
+            else self.original_file.name
+        )
+
+    def get_processing_status_text(self) -> str:
+        status = [
```
gecBurton, line 83: nit pick: could use `next` here

django_app/redbox_app/redbox_core/views.py
```diff
@@ -141,30 +133,28 @@ def upload_view(request):
...
 
 @login_required
 def sessions_view(request, session_id: str = ""):
-    
     USE_STREAMING = False
     STREAMING_ENDPOINT = "ws://localhost:8888"
-    
-    chat_history = ChatHistory.objects.all().filter(users=request.user)
+
+    chat_history = models.ChatHistory.objects.all().filter(users=request.user)
```
gecBurton, line 153:
```suggestion
    chat_history = models.ChatHistory.objects.filter(users=request.user)
```

django_app/redbox_app/redbox_core/models.py
```diff
@@ -55,22 +55,37 @@ class ProcessingStatusEnum(models.TextChoices):
...
-    original_file = models.FileField(storage=settings.BUCKET_NAME)
+    processing_status = models.CharField(
+        choices=ProcessingStatusEnum.choices, null=False, blank=False
+    )
+    original_file = models.FileField(storage=settings.DEFAULT_FILE_STORAGE)
     user = models.ForeignKey(User, on_delete=models.CASCADE)
+    original_file_name = models.TextField(max_length=2048, blank=True, null=True)
 
     @property
     def file_type(self):
```
gecBurton, line 66:
```suggestion
    def file_type(self) -> str:
```

django_app/redbox_app/redbox_core/models.py
```diff
@@ -55,22 +55,37 @@ class ProcessingStatusEnum(models.TextChoices):
...
 
     @property
-    def file_url(self):
+    def url(self):
         return self.original_file.url
 
     @property
     def name(self):
-        return self.original_file.name
+        return (
```
gecBurton, line 76: the logic here is confusing me, what if `self.original_file_name` was `None` , `""` ?

django_app/redbox_app/redbox_core/views.py
```diff
@@ -127,6 +112,13 @@ def upload_view(request):
 
             try:
                 api.upload_file(uploaded_file.name, request.user)
+                file = models.File.objects.create(
```
gecBurton, line 115: does this ORM code to be outside of this try/except?

django_app/redbox_app/redbox_core/views.py
```diff
@@ -189,21 +179,21 @@ def post_message(request: HttpRequest) -> HttpResponse:
...
     # save user message
-    chat_message = ChatMessage(chat_history=session, text=message_text, role=ChatRoleEnum.user)
+    chat_message = models.ChatMessage(chat_history=session, text=message_text, role=models.ChatRoleEnum.user)
     chat_message.save()
 
     # get LLM response
     message_history = [
         {"role": message.role, "text": message.text}
-        for message in ChatMessage.objects.all().filter(chat_history=session)
+        for message in models.ChatMessage.objects.all().filter(chat_history=session)
```
gecBurton, line 196:
```suggestion
        for message in models.ChatMessage.objects.filter(chat_history=session)
```

django_app/redbox_app/redbox_core/models.py
252afh: The names probably aren't helping tbf.

If django-storages sees a file with a duplicate name it will append random strings to the end. I'm storing the original name separately so it gets saved before django changes it in the filefield. This way people won't see `my_file_QWLSOX.pdf` in the list.

The check is to show the original name before showing the django-set file name, but with backwards compatibility for files that don't have it set yet in the db

django_app/redbox_app/redbox_core/views.py
252afh: I think it will stay inside as the except just adds errors. If I move it after+outside the try/except then if the upload fails django will still save the file. If the upload fails then the saving should fail too

gecBurton (approved):
some nitpcking added which can be ignored

otherwise all nicely written and clear, could do with some tests though

tested locally, all works 👍

PR #314, opened 2024-05-07
Feature/streaming

gecBurton:
## Context

This is a POC to investigate:
* how streaming works in langchain
* if/how websockets integrate with fastapi
* how chat-requests and responses might work with websockets

I have not:
* tested this code
* given any thought to security
* considered how to limit the document retrieval to just those owned by the requestor
* integrated this with the existing rag prompts/code

## Guidance to review

I have included a small http page to demonstrate this working, note that the chat is sent as text and the documents as binary, although this is just the first thing i thought of an not necessarily the best way to do this.

`docker compose up core-api elasticsearch`

http://0.0.0.0:5002/chat/chit-chat

image

## Things to check


KevinEtchells:
Thanks for your work on this @gecBurton! I'm not able to test locally (general issue running core-api, not to do with this work), but it looks promising.

Thinking in terms of how this looks to the user client-side, I think it'd make sense for the text to be streamed, and the documents to be sent all together as one chunk. I think Django should handle this though, so core-api just streams away. Not sure if using binary for the documents affects this. @brunns do you have any thoughts on this?

gecBurton:
> Thanks for your work on this @gecBurton! I'm not able to test locally (general issue running core-api, not to do with this work), but it looks promising.
> 
> Thinking in terms of how this looks to the user client-side, I think it'd make sense for the text to be streamed, and the documents to be sent all together as one chunk. I think Django should handle this though, so core-api just streams away. Not sure if using binary for the documents affects this. @brunns do you have any thoughts on this?

@KevinEtchells It appears that the documents are the first object to be sent back and all in one go (which i guess makes sense as the `R` needs to be done before the `AG`?). Binary streams in the same way as text. But it might make more sense to wrap them in and identifier, i.e.:

```json
{"resource_type": "documents", "data": [...]}
```

```json
{"resource_type": "text", "data": "As a large language model I cannot..."}
```

KevinEtchells:
> > Thanks for your work on this @gecBurton! I'm not able to test locally (general issue running core-api, not to do with this work), but it looks promising.
> > Thinking in terms of how this looks to the user client-side, I think it'd make sense for the text to be streamed, and the documents to be sent all together as one chunk. I think Django should handle this though, so core-api just streams away. Not sure if using binary for the documents affects this. @brunns do you have any thoughts on this?
> 
> @KevinEtchells It appears that the documents are the first object to be sent back and all in one go (which i guess makes sense as the `R` needs to be done before the `AG`?). Binary streams in the same way as text. But it might make more sense to wrap them in and identifier and stream both as text, i.e.:
> 
> ```json
> {"resource_type": "documents", "data": [...]}
> ```
> 
> ```json
> {"resource_type": "text", "data": "As a large language model I cannot..."}
> ```
> 
> this way the client can deal with the data as it chooses?

Great that documents all comes all in one go, and it makes sense in terms of what you have available first. We'll likely be displaying the resources last and therefore want to prioritise getting the text in as soon as possible, but whether sending documents first actually has any noticeable affect on performance would need to be tested. So probably best to stick to this default behaviour for now.

brunns:
Just checking - we are going to be using a json structure here, even though it'll be coming over in chunks?

gecBurton:
> Just checking - we are going to be using a json structure here, even though it'll be coming over in chunks?

@brunns the question i am trying to address with json is "as a client how do i know if the chunk i have just received is a document or a piece of text?"

brunns:
I see the issue. Trouble is, partial JSON is a real pain to deal with too. How about; firstly, documents are sent chunk by chunk, followed by a known delimiter, followed by AI messages.

KevinEtchells:
> I see the issue. Trouble is, partial JSON is a real pain to deal with too. How about; firstly, documents are sent chunk by chunk, followed by a known delimiter, followed by AI messages.

That's a good point. We want to avoid JSON for the main text. Documents less critical (especially if coming in one chunk), but I think @brunns suggestion is a good one.

gecBurton:
> > I see the issue. Trouble is, partial JSON is a real pain to deal with too. How about; firstly, documents are sent chunk by chunk, followed by a known delimiter, followed by AI messages.
> 
> That's a good point. We want to avoid JSON for the main text. Documents less critical (especially if coming in one chunk), but I think @brunns suggestion is a good one.

I was thinking that the chunks themselves could be json encoded, i.e.:

## chunk 1

```json
{"resource_type": "text", "data": "As a large language model"}
```

## chunk 2
```json
{"resource_type": "text", "data": " I cannot comment"}
```

## chunk 3
{"resource_type": "documents", "data": [{"url": "s3://amazon.sts/redboxdata/myDocyment.pdf"}]}
```

## chunk 4
```json
{"resource_type": "text", "data": " on whatever it is you just asked me."}
```

this way:
* nothing is partial
* it is always clear whether the chunk is:
   * The Documents
   * A piece of text

or we could stick to the binary=documents, text=chat distinction ?

brunns:
I'd be very happy with that. The text might come in multiple chunks, right?

gecBurton:
> I'd be very happy with that. The text might come in multiple chunks, right?

yep, almost certainly

core_api/src/routes/chat.py
```diff
@@ -162,3 +168,125 @@ def rag_chat(chat_request: ChatRequest, _user_uuid: Annotated[UUID, Depends(get_
...
+@chat_app.websocket("/rag-stream")
+async def websocket_endpoint(websocket: WebSocket):
+    await websocket.accept()
+
+    prompt_search_query = ChatPromptTemplate.from_messages(
+        [
+            MessagesPlaceholder(variable_name="chat_history"),
+            ("user", "{input}"),
+            (
+                "user",
```
rachaelcodes, line 182: Is this a system prompt?

rachaelcodes: (this question might be moot when integrating with existing code from redbox/llm/prompts/chat.py )

core_api/src/routes/chat.py
```diff
@@ -162,3 +168,125 @@ def rag_chat(chat_request: ChatRequest, _user_uuid: Annotated[UUID, Depends(get_
...
+        ]
+    )
+
+    document_chain = create_stuff_documents_chain(llm, prompt_get_answer)
+    retrieval_chain = create_retrieval_chain(retriever_chain, document_chain)
+
+    while True:
+        chat_request = ChatRequest.parse_raw(await websocket.receive_text())
+        chat_history = [
+            HumanMessage(content=x.text) if x.role == "user" else AIMessage(content=x.text)
```
rachaelcodes, line 204: Can there be `SystemMessage`s or will they have been contracted by the chains?

rachaelcodes:
An interesting UX question - do we want users to be able to see the generated question with their responses? (e.g. 'Does Rishi Sunak have a cat?' in George's example) Perhaps it could be something that could be revealed behind a click?

Or would it be better to show just the questions they wrote, the answers and the sources?

core_api/src/routes/chat.py
```diff
@@ -162,3 +168,125 @@ def rag_chat(chat_request: ChatRequest, _user_uuid: Annotated[UUID, Depends(get_
...
+
+            // Change binary type from "blob" to "arraybuffer"
+            ws.binaryType = "arraybuffer";
+
+            var decoder = new TextDecoder("utf-8");
+
+            // Listen for messages
+            ws.addEventListener("message", (event) => {
+              if (event.data instanceof ArrayBuffer) {
+                document.getElementById('documents').innerHTML = JSON.stringify(JSON.parse(decoder.decode(event.data)),null,2);
```
rachaelcodes, line 244: Do we want to be keeping a record of the sources throughout the chat rather than just the most recent?  If so, would this be embedded with the relevant chat answer or all in one block at the end?

core_api/src/routes/chat.py
```diff
@@ -162,3 +168,125 @@ def rag_chat(chat_request: ChatRequest, _user_uuid: Annotated[UUID, Depends(get_
...
+        chat_history = [
+            HumanMessage(content=x.text) if x.role == "user" else AIMessage(content=x.text)
+            for x in chat_request.message_history[:-1]
+        ]
+        chat = {"chat_history": chat_history, "input": chat_request.message_history[-1].text}
+
+        async for event in retrieval_chain.astream_events(chat, version="v1"):
+            kind = event["event"]
+            if kind == "on_chat_model_stream":
+                await websocket.send_text(event["data"]["chunk"].content)
```
brunns, line 212: Is this where we'll be wrapping the chunk on JSON?

gecBurton: yep, i guess so

brunns: Happy for me to jump in and take a run at it this afternoon?

gecBurton: Can we wait until after planning? I want to break this one down a bit and farm out the pieces rather than tackle this in one go

brunns: Sounds sensible.

django_app/tests/test_consumers.py
```diff
@@ -1,92 +1,92 @@
...
-        "POST", rag_url, json={"output_text": "Good afternoon, Mr. Amor.", "source_documents": []}
-    )
+    mocked_connect: Connect = create_mocked_connect(uploaded_file)
 
     # When
-    communicator = WebsocketCommunicator(ChatConsumer.as_asgi(), "/ws/chat/")
-    communicator.scope["user"] = carol
-    connected, subprotocol = await communicator.connect()
-    assert connected
+    with patch("redbox_app.redbox_core.consumers.connect", new=mocked_connect):
```
gecBurton, line 57: :+1:

brunns (approved):

PR #317, opened 2024-05-07
Django layer of chat streaming

brunns:
## Things to check


redbox/models/settings.py
```diff
@@ -78,6 +78,8 @@ class Settings(BaseSettings):
     from_email: Optional[str] = None
     email_file_path: str = "/app/mail"
     govuk_notify_plain_email_template_id: str = "example-id"
+    use_streaming: bool = False
```
gecBurton, line 81: :+1:

.env.example
```diff
@@ -67,3 +67,5 @@ NOTIFY_API_KEY=f4k3_k3y
 FROM_EMAIL=test@example.com
 GOVUK_NOTIFY_PLAIN_EMAIL_TEMPLATE_ID=example-id
 EMAIL_FILE_PATH='/app/mail'
+USE_STREAMING=False
```
rachaelcodes, line 70: Is this only used by the Django app? Maybe we don't need the new env variable in the non-Django environment list?

django_app/redbox_app/routing.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from .redbox_core import consumers
+
+websocket_urlpatterns = [
+    path(r"ws/chat/", consumers.ChatConsumer.as_asgi(), name="chat"),
```
rachaelcodes, line 6: are we hardcoding this as well as using an environment variable for it?

.env.example
brunns: It is. I didn't know you could set the env variables anywhere else! Where would a Django specific setting go?

django_app/redbox_app/routing.py
brunns: Good point.I don't actually think it needs to be configurable - none of the other paths are. I'll tear that out.

django_app/redbox_app/redbox_core/views.py
```diff
@@ -155,21 +155,18 @@ def remove_doc_view(request, doc_id: str):
...
-    STREAMING_ENDPOINT = "ws://localhost:8888"
-
+def sessions_view(request: HttpRequest, session_id: str = ""):
     chat_history = ChatHistory.objects.all().filter(users=request.user)
 
     messages = []
     if session_id:
         messages = ChatMessage.objects.filter(chat_history__id=session_id)
-
+    endpoint = URL.build(scheme="ws", host=request.get_host(), path=settings.STREAMING_ENDPOINT)
```
rachaelcodes, line 164: Should the endpoint change depending on whether settings.USE_STREAMING is true or not?

brunns: If streaming is False, we don't use the endpoint at all.

.env.example
rachaelcodes: Hm what this is showing me is that our dev docs are not very clear on the purpose of the 4 different sample .env files. `.env.django` is specifically for settings for running the application with Django (https://github.com/i-dot-ai/redbox-copilot/blob/cab97583c7a4018fc8c07a6b9078e153980a49eb/.github/workflows/django-checks.yml#L36) but maybe my question is moot given that we're thinking of streamlining our env files.

django_app/redbox_app/redbox_core/views.py
rachaelcodes: Maybe it's worth a comment explaining that?

rachaelcodes (approved):
Just a couple of questions about environment things I wasn't sure about, but looks good and the tests are clear.

PR #326, opened 2024-05-09
Feature/convert local storage back to minio

252afh:
## Context

Local storage should use minio instead of localfile because it will make compatibility with how core etc does files

- Set up local settings to use boto3 with minio instead of local file storage
- Use furl for urls

## Guidance to review

- Launch the app locally
- Upload a file
- Try to download a file
- Check localhost:9000 (minio) for the file
- Try to delete the file

## Things to check


252afh:
This PR is going into another branch, not `main`, so we can check the changes incrementally

.env.django
```diff
@@ -35,7 +35,21 @@ INGEST_QUEUE_NAME=redbox-ingester-queue
 REDIS_HOST=redis
 REDIS_PORT=6379
 
+# === Object Storage ===
```
252afh, line 38: re-adding and slightly modifying settings in the .env files

django_app/redbox_app/redbox_core/client.py
```diff
@@ -1,9 +1,33 @@
+import boto3
```
252afh, line 1: re-added s3 client for tests to run

django_app/pyproject.toml
```diff
@@ -43,6 +43,7 @@ django-allauth = "^0.61.1"
 django-otp = "^1.3.0"
 humanize = "^4.9.0"
 yarl = "^1.9.4"
+furl = "^2.1.3"
```
252afh, line 46: re-added and updated version of furl

django_app/redbox_app/redbox_core/migrations/0004_alter_file_original_file.py
```diff
@@ -1,6 +1,4 @@
 # Generated by Django 4.2.11 on 2024-05-03 10:08
```
252afh, line 1: remove need for checking for storage as both local and deployed are using boto3

django_app/pyproject.toml
brunns: Sorry - I think this was my mistake - but furl is a huge foot-pistol. yarl is far, far better better, because its URL instances are immutable.

brunns (approved):

PR #335, opened 2024-05-10
Make logging show up from Django app running in container.

brunns:
## Things to check

### DIFF
--- a/django_app/redbox_app/settings.py
+++ b/django_app/redbox_app/settings.py
@@ -236,24 +236,26 @@ DATABASES = {
     }
 }
 
+LOG_LEVEL = "DEBUG"
 LOGGING = {
     "version": 1,
     "disable_existing_loggers": False,
     "formatters": {"verbose": {"format": "%(asctime)s %(levelname)s %(module)s: %(message)s"}},
     "handlers": {
         "file": {
-            "level": "DEBUG",
+            "level": LOG_LEVEL,
             "class": "logging.FileHandler",
             "filename": os.path.join(LOG_ROOT, "application.log"),
             "formatter": "verbose",
         },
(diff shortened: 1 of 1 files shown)


KevinEtchells:
Hi @brunns, I can see the additional logging info and I'm able to log custom messages - great!
Just wondering if DEBUG is too verbose - would INFO suffice, or are you specifically needing some of the debug messages?

brunns:
> Hi @brunns, I can see the additional logging info and I'm able to log custom messages - great! Just wondering if DEBUG is too verbose - would INFO suffice, or are you specifically needing some of the debug messages?

Yeah, debug messages are **exactly** the ones I'm looking to see. How would you feel about this?

```python
LOG_LEVEL = "DEBUG" if DUBUG else "INFO"
```

brunns:
Or we could pick the log level straight out of `.env`?

KevinEtchells:
Either of those work for me - do you have a preference? Happy to approve with whichever approach you prefer.

brunns:
The latter is more flexible, so let's go with that.

.env.example
```diff
@@ -54,6 +54,7 @@ REDIS_PORT=6379
 DJANGO_SETTINGS_MODULE=redbox_app.settings
 DEBUG=True
 DJANGO_SECRET_KEY=1n53cur3K3y
+DJANGO_LOG_LEVEL="DEBUG"
```
KevinEtchells, line 57: Is this one meant to be different to the other .envs?

brunns: Yup - I thought that this is the ones devs start with, and they are likely to want to see debug messages.

KevinEtchells (approved):

PR #361, opened 2024-05-14
REDBOX 225 - display link to sources

brunns:
## Context

As per https://technologyprogramme.atlassian.net/browse/REDBOX-225 - display link to sources.

When uploading documents to the Core API, we now store the UUID it returns. We then use that to retrieve the relevant File objects, and add them to the ChatMessage.

Note that we still need to display them properly.

## Things to check


django_app/redbox_app/redbox_core/views.py
```diff
@@ -180,12 +179,16 @@ def post_message(request: HttpRequest) -> HttpResponse:
...
 
-    # save LLM response
-    llm_message = ChatMessage(chat_history=session, text=output_text, role=ChatRoleEnum.ai)
+    llm_message = ChatMessage(chat_history=session, text=response_data.output_text, role=ChatRoleEnum.ai)
     llm_message.save()
 
+    files: list[File] = [
+        File.objects.get(core_file_uuid=doc.file_uuid, user=request.user) for doc in response_data.source_documents
+    ]
+    llm_message.source_files.set(files)
```
brunns, line 190: Every other change here is just to make this work.

django_app/redbox_app/redbox_core/models.py
```diff
@@ -59,6 +59,10 @@ class File(UUIDPrimaryKeyBase, TimeStampedModel):
     original_file = models.FileField(storage=settings.STORAGES["default"]["BACKEND"])
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     original_file_name = models.TextField(max_length=2048, blank=True, null=True)
+    core_file_uuid = models.UUIDField(null=False)
```
gecBurton, line 62:
```suggestion
    core_file_uuid = models.UUIDField(null=True, blank=True)
```
otherwise this migration will fail on exiting `File` records

django_app/redbox_app/redbox_core/models.py
```diff
@@ -59,6 +59,10 @@ class File(UUIDPrimaryKeyBase, TimeStampedModel):
...
+
+    class Meta:
+        unique_together = ["user", "core_file_uuid"]
```
gecBurton, line 65:
```suggestion
        unique_together = ["id", "core_file_uuid"]
```
isnt it these two that are "the same"?

django_app/redbox_app/redbox_core/views.py
```diff
@@ -99,21 +98,21 @@ def upload_view(request):
...
-                    original_file_name=uploaded_file.name,
-                )
-                file.save()
-                # TODO: update improved File object with elastic uuid
-            except ValueError as value_error:
-                errors.append(value_error.args[0])
-
-            try:
-                api.upload_file(settings.BUCKET_NAME, uploaded_file.name, request.user)
+                upload_file_response = api.upload_file(settings.BUCKET_NAME, uploaded_file.name, request.user)
```
gecBurton, line 101: `api.upload_file` doesnt really upload a file it ingests an existing file that has already been uploaded (this should really be renamed!), so this must be called after `file.save()` ?

KevinEtchells:
Thanks @brunns. What would be really helpful is if you can log the filenames and file-urls inside the view (for now). That will help me to know what to do when adding to the template. Is that possible?

django_app/redbox_app/redbox_core/views.py
```diff
@@ -180,12 +179,16 @@ def post_message(request: HttpRequest) -> HttpResponse:
...
     core_api = CoreApiClient(host=settings.CORE_API_HOST, port=settings.CORE_API_PORT)
-    output_text = core_api.rag_chat(message_history, request.user.get_bearer_token())
+    response_data = core_api.rag_chat(message_history, request.user.get_bearer_token())
 
-    # save LLM response
-    llm_message = ChatMessage(chat_history=session, text=output_text, role=ChatRoleEnum.ai)
+    llm_message = ChatMessage(chat_history=session, text=response_data.output_text, role=ChatRoleEnum.ai)
     llm_message.save()
 
+    files: list[File] = [
```
gecBurton, line 187:
```python
files: list[File] = File.objects.filter(core_file_uuid__in=[doc.file_uuid or doc in response_data.source_documents], user=request.user)
```

this will be a fair bit faster

django_app/redbox_app/redbox_core/client.py
```diff
@@ -41,24 +47,24 @@ def __init__(self, host: str, port: int):
...
-            return file
+            return SimpleNamespace(key=name, bucket=bucket_name, uuid=uuid.uuid4())
 
         response = requests.post(
             f"{self.url}/file", json={"key": name}, headers={"Authorization": user.get_bearer_token()}, timeout=30
         )
         if response.status_code != 201:
             raise ValueError(response.text)
-        return response.json()
+        return response.json(object_hook=lambda d: SimpleNamespace(**d))
```
gecBurton, line 59: TIL. v nice :)

gecBurton (changes requested):
pls change:
1. https://github.com/i-dot-ai/redbox-copilot/pull/361/files#r1600207870

everything else is just nitpicking

django_app/redbox_app/redbox_core/models.py
brunns: There are two migrations. The first adds a default value, the second make the field non-nullable, so it should work?

django_app/redbox_app/redbox_core/models.py
brunns: Nope - the `id` is a UUID, the `user` is a foreign key.

django_app/redbox_app/redbox_core/views.py
brunns: Oh, I misunderstood. I'll look into that.

brunns: Ah - we'll need to discuss. If we haven't hit the upload file endpoint, we don't know the core API UUID.

django_app/redbox_app/redbox_core/views.py
brunns: Fixed.

django_app/redbox_app/redbox_core/models.py
gecBurton: afraid not, you will need an intermediate migration to populate the missing values

django_app/redbox_app/redbox_core/models.py
gecBurton: yes but isnt it the files that are supposed to match rather than a user and a file?

django_app/redbox_app/redbox_core/views.py
gecBurton: ah good point

django_app/redbox_app/redbox_core/models.py
gecBurton: I think we want:
* the `core_file_uuid` to be unique (in its own right)
* a 1:1 relationship between the original real file (`id`) and its embedded file (`core_file_uuid`)
?

gecBurton: I think you need:
1. a migration to add a nullable field with a default value
2. a data migration to back populate the new field for all the existing records 
3. a migration to make the new field non nullable

it is (2) that is missing, without it (3) will fail no?

brunns: Take a look now?

django_app/redbox_app/redbox_core/admin.py
```diff
@@ -2,5 +2,12 @@
 
 from . import models
 
+
+class FileResource(admin.ModelAdmin):
```
gecBurton, line 6: :+1:

good to have this, i always forget

252afh (approved):

PR #365, opened 2024-05-14
Redirect to documents page on sucessful file upload

KevinEtchells:
## Context

Currently, on successful file upload, it directs back to the document upload page. This then causes problems with the success message appearing when it shouldn't

* Redirect to documents page on successful file upload
* Remove back links (these weren't in the correct containers, but they're not needed as there are cancel buttons)

## Guidance to review

* Check a successful file upload
* Check uploading an unsupported file still stays on the same page and shows the error

### DIFF
--- a/django_app/redbox_app/templates/remove-doc.html
+++ b/django_app/redbox_app/templates/remove-doc.html
@@ -5,7 +5,6 @@
 
 {% block content %}
 
-<a href="{{url('documents')}}" class="govuk-back-link">Back</a>
 <form method="post" enctype="multipart/form-data">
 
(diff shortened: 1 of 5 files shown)


django_app/redbox_app/templates/remove-doc.html
```diff
@@ -5,7 +5,6 @@
 
 {% block content %}
 
-<a href="{{url('documents')}}" class="govuk-back-link">Back</a>
```
252afh, line 8: Leave the back links in, I think it's GDS standards to have them

252afh (approved):

PR #368, opened 2024-05-15
Bugfix/sort out urls

252afh:
## Context

URL's in templates and generally the flow is a bit borked

- Used jinja2 url pipe throughout the templates
- Added check for a logged-in user trying to get to logged-out views
- Broke up urls.py into more logical groups
- Looks like at some point documents.html was reverted to an earlier version, so updates to that have been re-applied

## Guidance to review

- Navigate each page and check that the links work
- Check that you can't login twice

## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/auth_views.py
+++ b/django_app/redbox_app/redbox_core/auth_views.py
@@ -11,13 +11,15 @@ logger = logging.getLogger(__name__)
 
 
 def sign_in_view(request: HttpRequest):
+    if request.user.is_authenticated:
+        return redirect("homepage")
     if request.method == "POST":
         form = SignInForm(request.POST)
         if form.is_valid():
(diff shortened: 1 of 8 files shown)


django_app/redbox_app/redbox_core/auth_views.py
```diff
@@ -11,13 +11,15 @@
 
 
 def sign_in_view(request: HttpRequest):
+    if request.user.is_authenticated:
```
gecBurton, line 14: we should add this to the test (later)

gecBurton (approved):
LGTM

django_app/redbox_app/redbox_core/auth_views.py
252afh: Agreed 👍

PR #369, opened 2024-05-15
Add condition to use small instance type for non prod env.

dk-singh:
## Context

The instance type of postgres is defined as `db.t3.large` for all prod and non-prod environment. The dev should be using the smaller instance to save cost. 

Add a condition based on environment to use small instance for non prod environment. 

## Relevant links

- https://github.com/i-dot-ai/i-ai-core-infrastructure/issues/251

## Things to check


dk-singh:
Thanks for your review 🙇‍♂️

PR #374, opened 2024-05-15
feature/REDBOX-252-failure-to-embed-doc-leads-to-a-poor-error-message

gecBurton:

If the core-api failed for any reason, this reason was passed back to the user. 

image

Now the raw error it is logged, and the user sees something more useful.

image

## Things to check


KevinEtchells (approved):
This message is much clearer. I wonder if users will understand what core-api is - could be "failed to connect to server"? But I'm not certain on that, and this is a good step forwards so approved!

PR #375, opened 2024-05-15
Add table support for markdown to HTML conversion

KevinEtchells:
## Context

We need to support markdown tables when converting to HTML

It now looks like this:

image

## Guidance to review

Generate a response with a table. See how it looks.


gecBurton (approved):
love it

PR #376, opened 2024-05-15
Feature/document status check

252afh:
## Context

Documents being uploaded should update their status as it goes through the system.

- Use polling from JS to get django to poll core for info
- Altered the JS to use the json response
- Added client call for getting file status from core
- Added a view to django to get file_status from core or try use s3 file status if present or return unknown

## Guidance to review

- Go to documents page
- Upload a document
- See the status change in the file list

To test error states
- Change the ID used in the file status endpoint to send the file id instead of the core file id
- See that the file status changes to `unknown`

## Things to check


brunns (commented):
Looks good. Might be worth rebasing over main and using the `core_file_uuid` from the File model now we are collecting it, and commit when it's fully working.

django_app/redbox_app/redbox_core/types.py
```diff
@@ -0,0 +1,21 @@
+from typing import Optional
+from uuid import UUID
+
+from redbox_app.redbox_core import models
+
+
+class ChunkStatus:
```
gecBurton, line 7: feels like DRF?

252afh: I'm 50/50 on adding the extra weight that adding DRF will do, I'm hoping that this will be the only non-view url in the django app

gecBurton: fair point

252afh: Fully agree though that if we even think we'll add more in the future then we should add DRF, I'll raise it at tomorrows stand-up

django_app/redbox_app/redbox_core/views.py
```diff
@@ -104,7 +112,7 @@ def upload_view(request):
     )
 
 
-def injest_file(uploaded_file: UploadedFile, user: User) -> list[str]:
+def ingest_file(uploaded_file: UploadedFile, user: User) -> list[str]:
```
brunns, line 115: Doh! Sorry about that.

django_app/redbox_app/redbox_core/views.py
```diff
@@ -200,6 +209,33 @@ def post_message(request: HttpRequest) -> HttpResponse:
...
+@require_http_methods(["GET"])
+@login_required
+def file_status_api_view(request: HttpRequest) -> JsonResponse:
+    file_id = request.GET.get("id", None)
+    if not file_id:
+        logger.error("Error getting file object information - no file ID provided %s.")
+        return JsonResponse({"status": ProcessingStatusEnum.unknown.label})
+    try:
+        file = File.objects.get(pk=file_id)
+    except ObjectDoesNotExist as ex:
```
brunns, line 221: Perhaps `File.DoesNotExist`? I generally prefer narrow exceptions over broad ones.

django_app/redbox_app/redbox_core/client.py
```diff
@@ -67,3 +67,10 @@ def rag_chat(self, message_history: list[dict[str, str]], token: str) -> SimpleN
         logger.debug("response_data: %s", response_data)
 
         return response_data
+
+    def get_file_status(self, file_id: uuid, token: str) -> SimpleNamespace:
```
brunns, line 71: Other client methods take the User object and pull the token out of it rather than just passing the token. It might be worth being consistent.

brunns (commented):
Post MVP it might be worth considering a streaming approach here rather than a polling one. Most of the pieces are in place for this on the Django side, put there for the streaming chat.

django_app/redbox_app/redbox_core/client.py
```diff
@@ -67,3 +67,10 @@ def rag_chat(self, message_history: list[dict[str, str]], token: str) -> SimpleN
...
+        url = f"{self.url}/file/{file_id}/status"
```
brunns, line 72: Could be `self.url / "file" / file_id / "status"`

django_app/redbox_app/redbox_core/views.py
gecBurton: i think this was my fault

django_app/redbox_app/redbox_core/client.py
252afh: Yeah I'm going to raise a ticket for tidying up the client, it could do with a bit of refinement

brunns (approved):

PR #377, opened 2024-05-15
Handle duplicate file names

rachaelcodes:
## Context

Previously, we didn't have a way to handle duplicate file names, as S3 as a default overwrites files with duplicate keys. 

This PR updates the `AWS_S3_FILE_OVERWRITE` setting (as in this discussion: https://github.com/jschneier/django-storages/issues/77 ). This causes S3 to append a hash onto duplicate filenames, allowing for multiple versions. 

I've also added some tests to confirm this behaves as expected.

Screenshot 2024-05-15 at 15 12 32

## Guidance to review

Try and upload the same file a couple of times. Confirm two different file names appear in `/documents` , and that you can download each file.

## Relevant links
 https://technologyprogramme.atlassian.net/browse/REDBOX-241

## Things to check


django_app/redbox_app/redbox_core/views.py
```diff
@@ -108,22 +108,22 @@ def upload_view(request):
                 )
                 file.save()
                 # TODO: update improved File object with elastic uuid
-            except ValueError as value_error:
-                errors.append(value_error.args[0])
 
-            try:
-                api.upload_file(settings.BUCKET_NAME, uploaded_file.name, request.user)
+                api.upload_file(settings.BUCKET_NAME, file.name, request.user)
```
gecBurton, line 112: just for my understanding, there will be separate embedding for each duplicate file name?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -96,7 +96,7 @@ def url(self) -> URL:
 
     @property
     def name(self) -> str:
-        return self.original_file_name if self.original_file_name else self.original_file.name
+        return self.original_file.file.name  # handles duplicate names
```
252afh, line 99: Do we need the extra `.file.` nest here?

Does this mean in the documents page the users see the documents name with the hash on the end?

django_app/redbox_app/redbox_core/views.py
```diff
@@ -108,22 +108,22 @@ def upload_view(request):
...
             except ValueError as value_error:
                 errors.append(value_error.args[0])
 
             if not errors:
-                return redirect(documents_view)
-
+                return redirect(reverse(documents_view))
```
gecBurton, line 117: :+1:

django_app/tests/test_views.py
```diff
@@ -37,6 +43,31 @@ def test_upload_view(alice, client, file_pdf_path, s3_client):
...
+
+        client.force_login(bob)
+        response = client.post("/upload/", {"uploadDoc": f})
+
+        assert response.status_code == 302
+        assert response.url == "/documents/"
+
+        assert count_s3_objects(s3_client) == previous_count + 3
+
+        assert File.objects.order_by("-created_at")[0].name != File.objects.order_by("-created_at")[1].name
```
gecBurton, line 68: :+1:

gecBurton (approved):
super clear ❤️

django_app/redbox_app/redbox_core/models.py
rachaelcodes: `original_file.file.name` gives the name with the hash, `original_file.name` give the name without the hash. 

We need the first for core-api so that the embeddings are different (in case files have the same name but different content). But I agree we can use original_file.name for the user, will update this now.

django_app/redbox_app/redbox_core/views.py
rachaelcodes: There should be. The uniqueness of the `name` property is tested in the line: 
`assert File.objects.order_by("-created_at")[0].name != File.objects.order_by("-created_at")[1].name`

Based on Ell's comment, I'll create a new property (`unique_name`) for sharing to core_api, so that `name` is just the uploaded name the user has seen.

gecBurton: thx for the explanation :+1:

brunns (commented):
Do we have different line length settings somewhere. A few of these changes are formatting changes that linting from the command line isn't doing.

django_app/redbox_app/redbox_core/models.py
252afh: Ok that sounds ideal, if that's the case - that we can get both hashed and unhashed name from the file - then we probably don't need the property on the File model and can change the documents views to reflect that

252afh: (unless we want to make it more readable in usages, in which case I'm fine with the two properties)

rachaelcodes:
> Do we have different line length settings somewhere. A few of these changes are formatting changes that linting from the command line isn't doing.

This is what's coming out of `make format` on my machine. Sounds like we need your ruff formatting document :)

252afh (approved):

gecBurton:
> > Do we have different line length settings somewhere. A few of these changes are formatting changes that linting from the command line isn't doing.
> 
> This is what's coming out of `make format` on my machine. Sounds like we need your ruff formatting document :)

agree with @rachaelcodes here, 1 PR for duplicate file names and another for global formatting

brunns:
> > Do we have different line length settings somewhere. A few of these changes are formatting changes that linting from the command line isn't doing.
> 
> This is what's coming out of `make format` on my machine. Sounds like we need your ruff formatting document :)

Ah, perhaps these changes will disappear when you merge from main.

PR #386, opened 2024-05-16
feature/omit-empty-sources

gecBurton:
## Context

As a user I do not want an empty source section to be displayed in addition to the explicit sources in the payload

## Things to check


KevinEtchells (approved):
Thanks George. This is still a good step, so approved!

KevinEtchells (approved):

PR #388, opened 2024-05-16
Use document unique name for minio download link

252afh:
Document unique name is needed to generate the downloadable url locally so that the file can be found in minio


brunns (approved):
Works for me.

PR #399, opened 2024-05-17
environment specific elastic index

gecBurton:
## Context

As an infra engineer I want the indexes in our single elasticseach instance to include the environment name so that users data is not shared across environments 

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/9123932880

### DIFF
--- a/core_api/src/routes/file.py
+++ b/core_api/src/routes/file.py
@@ -51,7 +51,7 @@ file_publisher = FilePublisher(router.broker, env.ingest_queue_name)
 # === Storage ===
 
 es = env.elasticsearch_client()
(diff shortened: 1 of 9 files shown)


core_api/src/routes/file.py
```diff
@@ -51,7 +51,7 @@ def file_not_found_response(file_uuid: UUID) -> JSONResponse:
 # === Storage ===
 
 es = env.elasticsearch_client()
-storage_handler = ElasticsearchStorageHandler(es_client=es, root_index="redbox-data")
+storage_handler = ElasticsearchStorageHandler(es_client=es, root_index=f"redbox-data-{env.environment}")
```
gecBurton, line 54: this is the key change

PR #406, opened 2024-05-17
call core api delete from django delete

rachaelcodes:
## Context

Currently the frontend deletion of a file only deletes the file from the Django and S3 storage, but not the Elastic storage.

Deleting a file from the user interface propagates a call to core-api to delete the file and its chunks. If that API request fails, then the file will not delete from the Django database.

If the call to the API fails, the user will see this view: 
Screenshot 2024-05-20 at 12 31 30

## Guidance to review

Try deleting files from django.

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-256

## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/views.py
+++ b/django_app/redbox_app/redbox_core/views.py
@@ -156,12 +156,12 @@ def remove_doc_view(request, doc_id: uuid):
             core_api.delete_file(file.core_file_uuid, request.user)
         except HTTPError as e:
             logger.error("Error deleting file object %s.", file, exc_info=e)
-            errors.append("failed to connect to core-api")
+            errors.append("There was an error deleting this file")
 
         else:
             logger.info("Removing document: %s", request.POST["doc_id"])
             file.delete()
-        return redirect("documents")
+            return redirect("documents")
 
     return render(
         request,
--- a/django_app/redbox_app/redbox_core/views.py
+++ b/django_app/redbox_app/redbox_core/views.py
@@ -24,6 +24,7 @@ from requests.exceptions import HTTPError
 from yarl import URL
 
 logger = logging.getLogger(__name__)
+core_api = CoreApiClient(host=settings.CORE_API_HOST, port=settings.CORE_API_PORT)
 
 CHUNK_SIZE = 1024
 # move this somewhere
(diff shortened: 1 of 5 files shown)


django_app/tests/test_views.py
```diff
@@ -195,7 +238,6 @@ def test_post_message_to_new_session(alice: User, client: Client, requests_mock:
     assert response.status_code == HTTPStatus.FOUND
     assert "Location" in response.headers
     session_id = URL(response.url).parts[-2]
-    assert ChatMessage.objects.get(chat_history__id=session_id, role=ChatRoleEnum.user).text == "Are you there?"
```
rachaelcodes, line 198: hm not sure why this got deleted - will fix on Monday

django_app/redbox_app/redbox_core/views.py
```diff
@@ -149,14 +149,27 @@ def ingest_file(uploaded_file: UploadedFile, user: User) -> list[str]:
 @login_required
 def remove_doc_view(request, doc_id: uuid):
     file = File.objects.get(pk=doc_id)
+    errors: list[str] = []
+
     if request.method == "POST":
-        logger.info("Removing document: %s", request.POST["doc_id"])
-        file.delete()
+        api = CoreApiClient(host=settings.CORE_API_HOST, port=settings.CORE_API_PORT)
```
gecBurton, line 155: this could be declared just once, maybe as a constant?

brunns: Ultimately I'd like to inject it.

252afh (approved):
Looks good to me, could we change the file deletion setting to seconds instead of days for future proofing please?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -64,6 +69,7 @@ class File(UUIDPrimaryKeyBase, TimeStampedModel):
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     original_file_name = models.TextField(max_length=2048, blank=True, null=True)
     core_file_uuid = models.UUIDField(null=True)
+    expiry_date = models.DateField(default=get_default_expiry_date)
```
gecBurton, line 72: either:
* this needs to be nullable or we will get an error on migration, or
* we could compute this field on the fly from the `created_at: datetime`
?

django_app/redbox_app/redbox_core/views.py
```diff
@@ -24,6 +24,7 @@
 from yarl import URL
 
 logger = logging.getLogger(__name__)
+core_api = CoreApiClient(host=settings.CORE_API_HOST, port=settings.CORE_API_PORT)
```
gecBurton, line 27: 🙏

gecBurton (changes requested):
one RFC, otherwise LGTM

gecBurton (approved):

PR #409, opened 2024-05-17
Document how to save and load dumps from the vector store

ghost:
## Context

For LLM evaluation we need:

* Versioned evaluation datasets, which references files and chunks the vector store
* Versioned snapshots of the vector store that contains those referenced files and chunks

This PR adds documentation on how to save and load snapshots of the vector store so we can study static versions of it. I have tested this code, and successfully saved and loaded snapshots of the vector store.

## Guidance to review

* Node devs: how bad is my node advice

## Relevant links

* [elasticsearch-dump](https://github.com/elasticsearch-dump/elasticsearch-dump)

## Things to check

- [x] I have run integration tests -- NA

### DIFF
--- a/docs/DEVELOPER_SETUP.md
+++ b/docs/DEVELOPER_SETUP.md
@@ -131,23 +131,23 @@ Notebooks with some standard methods to evaluate the LLM can be found in the [no
 
 You may want to evaluate using versioned datasets in conjunction with a snapshot of the pre-embedded vector store.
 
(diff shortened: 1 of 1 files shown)


KevinEtchells (approved):
The Node advice is good, @wpfl-dbt.

NVM is great for managing different versions of Node, but not essential if this is the only place it's going to be used, and it doesn't look like it requires a specific version other than latest LTS. But it's not bad advice to have it installed anyway!

PR #411, opened 2024-05-20
Feature/simple clusterer

gecBurton:
## Context

One of the main observations from using our own code is that:
1. embedding takes a long time on large documents, i.e. 45 minutes for the Iraq Summary
2. it is not clear that this embedding is still taking place

In this notebook I propose a simple and cheap way to cluster our text chunks such that there is much less to embed. Whilst the solution presented here is far from ideal it would be quick to implement and might improve the experience of our earliest users.

## Things to check


gecBurton:
> It seems to me that semantic chunking approaches like [this LangChain example](https://python.langchain.com/v0.1/docs/modules/data_connection/document_transformers/semantic-chunker/) and [this notebook](https://github.com/FullStackRetrieval-com/RetrievalTutorials/blob/main/tutorials/LevelsOfTextSplitting/5_Levels_Of_Text_Splitting.ipynb) all tend to rely on statistics computed using embeddings and cosine similarity. While I don't know this is the best in class way, this makes me think we should have a really good reason for deviating from this pattern, and we currently have no evaluation north star to do that. Because of this I wonder if using a quicker model to calculate embedding comparisons might be the better route for now. Based on [HuggingFace embedding benchmarks](https://huggingface.co/blog/mteb), perhaps a variety of MiniLM L6, [like this one](https://huggingface.co/nreimers/MiniLM-L6-H384-uncased).
> 
> If we did follow this approach, is `distance_threshold` a better approach than `n_clusters` here? We don't know how many clusters we want to form, but giving a threshold seems reasonable (and how many of the linked embedding approaches do it this way).

thx for the detailed feedback @wpfl-dbt , I simply dont have the expertise to comment on this but it sounds like you are right to stick to the standard approach.

PR #420, opened 2024-05-21
Feature/processing status is transient

gecBurton:
## Context

As a Engineer I want the process_status to be removed from the django app so that there is a single source of truth for it in the core-api

## Things to check


django_app/redbox_app/redbox_core/views.py
```diff
@@ -241,13 +239,8 @@ def file_status_api_view(request: HttpRequest) -> JsonResponse:
...
         logger.error("File object information from core not found - file does not exist %s.", file_id, exc_info=ex)
-        if not file.processing_status:
-            file.processing_status = ProcessingStatusEnum.unknown.label
-            file.save()
-        return JsonResponse({"status": file.processing_status})
-    file.processing_status = core_file_status_response.processing_status
-    file.save()
-    return JsonResponse({"status": file.get_processing_status_text()})
+        return JsonResponse({"status": ProcessingStatusEnum.unknown.label})
+    return JsonResponse({"status": core_file_status_response.processing_status})
```
gecBurton, line 243:
```suggestion
    return core_file_status_response
```
i would actually rather do this

PR #425, opened 2024-05-21
Feature/redbox 234 Add file expiry date

rachaelcodes:
## Context

Adds an expiry_date field for Django Files. This is initially set based on 30 days from File upload (or whatever you want to set the environment variable to), but can be amended later. For example, there is a later task to update the expiry date when a document is used in a RAG chat.

Creates an expiry_date field.
Updates the  model `save()` method to auto-populate this field in cases where there is not an existing expiry_date:

- if there is already a `created_at` date (because this is an existing pre-migration file), it adds the default time delta to this date
- if not (because this is a new file), it sets the date based on `timezone.now()`*

* I considered also setting this based on `created_at` in a post_save signal, but the Django docs discourage using signals due to unpredictability.

## Guidance to review

Run this branch locally, including migrations. Check in the Django shell to see if your documents have expiry dates added as expected. Also add new documents and see if they have expiry dates.

## Relevant links

## Things to check


django_app/redbox_app/redbox_core/models.py
```diff
@@ -64,10 +66,20 @@ class File(UUIDPrimaryKeyBase, TimeStampedModel):
...
 
     def __str__(self) -> str:  # pragma: no cover
         return f"{self.original_file_name} {self.user}"
 
+    def save(self, *args, **kwargs):
+        if not self.expiry_date:
+            if self.created_at:
+                #  Needed to populate the initial expiry date for existing Files
+                self.expiry_date = self.created_at + timedelta(seconds=settings.FILE_EXPIRY_IN_SECONDS)
+            else:
```
gecBurton, line 79: i dont think that the `created_at` is ever null?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -64,10 +66,20 @@ class File(UUIDPrimaryKeyBase, TimeStampedModel):
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     original_file_name = models.TextField(max_length=2048, blank=True, null=True)
     core_file_uuid = models.UUIDField(null=True)
+    expiry_date = models.DateTimeField(blank=True, null=True)
```
gecBurton, line 69: does this need to be persisted?

rachaelcodes: yes, because it gets updated when the file gets used by RAG

django_app/redbox_app/redbox_core/models.py
rachaelcodes: It is on the first save() of creating an object. I tried running `super.save()` before this, but it needs running again to store the expiry_date, which causes pk uniqueness issues. The other option is to use a signal, but as mentioned above, Django doesn't recommend it.

django_app/redbox_app/redbox_core/models.py
gecBurton: ah ofc

django_app/redbox_app/redbox_core/models.py
gecBurton: TIL. ok, this all makes sense

gecBurton (approved):
thx for the explanations - LGTM

PR #426, opened 2024-05-21
Rename Sessions to Chats

KevinEtchells:
## Context

Following user testing, it has been decided to rename "Sessions" to "Chats"

* Update text references
* Update code references where helpful and where it doesn't cause too much disruption

## Guidance to review

* Check all references (that a user would see) to "Session(s)" have been replaced with "Chat(s)"
* Check links etc. still work okay


gecBurton (approved):
LGTM

PR #428, opened 2024-05-21
Add Plausible Code

Samuel-Hoskin:
## Context

Adds Plausible code to enable basic analytics tracking. 

This introduces the code snippet with an if statement so we can separate analytics out per env  

https://technologyprogramme.atlassian.net/browse/REDBOX-220?atlOrigin=eyJpIjoiYmI5MTU3ZmVmZmRkNGQ4MjkyZDg4NWZkOTE0N2M2M2UiLCJwIjoiaiJ9


django_app/redbox_app/settings.py
```diff
@@ -147,11 +147,12 @@
 CSP_DEFAULT_SRC = (
     "'self'",
     "s3.amazonaws.com",
+    "plausible.io",
```
KevinEtchells, line 150: Do we need this in DEFAULT as well as SCRIPT?

KevinEtchells (approved):
Looks good. Just one question about something minor.

django_app/redbox_app/settings.py
Samuel-Hoskin: Yes just tried with out and it thew an CSP error

PR #435, opened 2024-05-22
Update file expiry with RAG chat

rachaelcodes:
## Context

File expiry dates are set on file creation (since #425 ), but should be updated to give another 30 days before expiry whenever they are quoted in a RAG chat.

On a call to `/post-message`, the returned source files each have their expiry date extended for another 30 days (or as long into the future as configured with env vars).

## Guidance to review

Could use Django shell to inspect a file expiry date before and after it is used in a RAG chat.

## Relevant links

## Things to check

- ~[ ] I have run integration tests~


django_app/redbox_app/redbox_core/views.py
```diff
@@ -222,6 +224,10 @@ def post_message(request: HttpRequest) -> HttpResponse:
     files: list[File] = File.objects.filter(core_file_uuid__in=doc_uuids, user=request.user)
     llm_message.source_files.set(files)
 
+    for file in files:
```
252afh, line 227: I _think_, according to core, this is just all files for that user, we don't need to use a response from core for that

django_app/tests/test_views.py
```diff
@@ -264,6 +270,10 @@ def test_post_message_to_existing_session(chat_history: ChatHistory, client: Cli
     assert (
         ChatMessage.objects.get(chat_history__id=session_id, role=ChatRoleEnum.ai).text == "Good afternoon, Mr. Amor."
     )
+    assert (
+        ChatMessage.objects.get(chat_history__id=session_id, role=ChatRoleEnum.ai).source_files.first() == uploaded_file
+    )
+    assert initial_file_expiry_date != File.objects.get(core_file_uuid=uploaded_file.core_file_uuid).expiry_date
```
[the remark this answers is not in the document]
rachaelcodes: Yeah, good idea :)

django_app/redbox_app/redbox_core/views.py
rachaelcodes: Is it not the ones returned as relevant by the LLM chain `core_api/src/routes/chat.py#L150-L151`?

brunns: I believe you're right, Rachael - it's just the "sources", i.e. the relevant files.

django_app/redbox_app/redbox_core/models.py
```diff
@@ -66,18 +66,18 @@ class File(UUIDPrimaryKeyBase, TimeStampedModel):
...
     core_file_uuid = models.UUIDField(null=True)
-    expiry_date = models.DateTimeField(blank=True, null=True)
+    last_referenced = models.DateTimeField(blank=True, null=True)
 
     def __str__(self) -> str:  # pragma: no cover
         return f"{self.original_file_name} {self.user}"
 
     def save(self, *args, **kwargs):
-        if not self.expiry_date:
+        if not self.last_referenced:
```
252afh, line 75: I wonder if we need this part if we have the migration? (or vice versa)

252afh (approved):

PR #443, opened 2024-05-22
Revert "bugfix/remove-sources-part-2"

gecBurton:
Reverts i-dot-ai/redbox-copilot#430


redbox/llm/prompts/chat.py
```diff
@@ -15,7 +15,7 @@
 
 
 _with_sources_template = """Given the following extracted parts of a long document and \
-a question, create a final answer.  \
+a question, create a final answer with Sources at the end.  \
```
brunns, line 18: I assume this puts the sources back?

brunns (approved):

PR #448, opened 2024-05-23
Feature/redbox 234 keep deleted file django reference

rachaelcodes:
## Context
After discussion with the wider product team about managing deleted documents, we decided to keep a reference to files in the Django database when the files were deleted from S3 and the core_api.

Now when a user chooses to delete an object, we delete the file from S3 rather than the whole Django object.

This also updates the UI with an interim version to show that the file is in a 'deleted' state.
Screenshot 2024-05-23 at 15 48 12

## Guidance to review

Run locally, including the database migrations (check nothing broken) - you might need to `make reset-db` before switching to a different branch. Try deleting a file and check how it displays - see if it deleted in minio.

## Relevant links

## Things to check

- ~[ ] I have run integration tests~


brunns (approved):
Looks good.

brunns:
Probably worth merging main in again and running `make format lint`.

PR #451, opened 2024-05-23
Adds parameter for security classification

ghost:
## Context

[REDBOX-294](https://technologyprogramme.atlassian.net/jira/software/projects/REDBOX/boards/409?assignee=712020%3A12d86952-e2fd-491d-97cc-c3a261385fe1&selectedIssue=REDBOX-294). Just as we allow many different LLMs to be used, different departments will permit different levels of security assurance on documents staff are permitted to use with Redbox. This ticket is to add an environment variable to allow the frontend to make changes based on an app-wide setting.

* Add `MAX_SECURITY_CLASSIFICATION`
* Add places for Django to pick it up and display it

## Guidance to review

* Checked my overall strategy with @KevinEtchells but make sure you're happy with how I pass that env around
* In particular, I use `case switch` because I think an extension of this kind of work would involve quite a few different settings that result from the max security level of documents permitted

## Relevant links

* [Government Security Classifications](https://www.gov.uk/government/publications/government-security-classifications)

## Things to check

[REDBOX-294]: https://technologyprogramme.atlassian.net/browse/REDBOX-294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


django_app/redbox_app/settings.py
```diff
@@ -323,3 +323,20 @@
 USE_STREAMING = env.bool("USE_STREAMING")
 FILE_EXPIRY_IN_SECONDS = env.int("FILE_EXPIRY_IN_DAYS") * 24 * 60 * 60
 SUPERUSER_EMAIL = env.str("SUPERUSER_EMAIL", None)
+
+# Security classifications
+# https://www.gov.uk/government/publications/government-security-classifications/
+
+MAX_SECURITY_CLASSIFICATION = env.str("MAX_SECURITY_CLASSIFICATION")
```
brunns, line 330: We could make this an enum:

```python
class Classification(enum.StrEnum):
    OFFICIAL = "Official"
    OFFICIAL_SENSITIVE = "Official Sensitive"
    ...
    
MAX_SECURITY_CLASSIFICATION = Classification[env.str("MAX_SECURITY_CLASSIFICATION")]
```

We'd need to use underscores rather than dashes in the names for them, of course.

django_app/redbox_app/jinja2.py
```diff
@@ -63,6 +63,7 @@ def environment(**options):
             "url": url,
             "humanize_timedelta": humanize_timedelta,
             "environment": settings.ENVIRONMENT,
+            "security": settings.MAX_SECURITY_REPR,
```
brunns, line 66: If we were to use an enum, we'd want `MAX_SECURITY_CLASSIFICATION.value`.

django_app/redbox_app/settings.py
brunns: Also - would we want a default value?

252afh: I'm fine withw/without an enum. r.e. default, we want it to fall over if they haven't set it

brunns (approved):
Looks good!

PR #487, opened 2024-05-30
bugfix/get-openai-working-again

gecBurton:
## Context

As a user I want to be able to use OpenAI as well as Azure

Previously we were passing `OPENAI_MODEL` into `ChatLiteLLM` (which by default is `azure/gpt-35-turbo-16k`), this is only required for azure so this envvar has been renamed `AZURE_OPENAI_MODEL`

## Things to check

### DIFF
--- a/core_api/src/routes/chat.py
+++ b/core_api/src/routes/chat.py
@@ -26,7 +26,7 @@ from redbox.llm.prompts.chat import (
 )
 from redbox.model_db import MODEL_PATH
 from redbox.models import EmbeddingModelInfo, Settings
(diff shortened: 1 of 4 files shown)


core_api/src/routes/chat.py
```diff
@@ -73,7 +73,6 @@ def populate_embedding_model_info() -> EmbeddingModelInfo:
 if env.openai_api_key is not None:
     log.info("Creating OpenAI LLM Client")
     llm = ChatLiteLLM(
-        model=env.openai_model,
```
gecBurton, line 76: this is the key change

lmwilkigov (approved):

PR #507, opened 2024-06-03
Added make command that will auto-deploy-to-dev by default, or any en…

252afh:
…v if specified

## Things to check


252afh:
Workflows only appear in the actions list once in the default branch, so merging to test in another branch

PR #508, opened 2024-06-03
fix prod deployment

gecBurton:
## Context

"redbox.ai.cabinetoffice.gov.uk" was not getting added to the ALLOWED_HOSTS because the ENVIRONMENT was case sensitive

## Things to check

### DIFF
--- a/django_app/redbox_app/settings.py
+++ b/django_app/redbox_app/settings.py
@@ -146,7 +146,7 @@ LOGIN_URL = "sign-in"
 
 HOST = (
     "redbox.ai.cabinetoffice.gov.uk"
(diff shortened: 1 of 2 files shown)


django_app/redbox_app/settings.py
```diff
@@ -146,7 +146,7 @@
 
 HOST = (
     "redbox.ai.cabinetoffice.gov.uk"
-    if ENVIRONMENT == "prod"
+    if ENVIRONMENT.lower() == "prod"
```
gecBurton, line 149: this is the key change

rachaelcodes (approved):

PR #512, opened 2024-06-04
Remove alpine.js

KevinEtchells:
## Context

We have stopped using Alpine.js due to it not playing nicely with our CSP. This is updating the final place where it's used.

* Changing `FileStatus` to a webcomponent
* Checking if element is visible before making a request (as different elements are used for mobile and desktop)

## Guidance to review

* Upload a doc and check the file status updates without needing a page reload


brunns (approved):
I can confirm it's still working st lerast.

PR #513, opened 2024-06-04
Feature/auto release action

252afh:
## Things to check


.github/workflows/manual-release.yml
```diff
@@ -48,48 +48,69 @@ jobs:
...
+          repository: i-dot-ai/redbox-copilot-infra-config
+          ref: refs/heads/main
+          token: ${{ secrets.AWS_GITHUBRUNNER_PAT }}
+          path: redbox-copilot-infra-config
+
+      - name: Make release
+        working-directory: redbox-copilot
+        run: |
+          make tf_init
+          make tf_auto_deploy env=${{ github.event.inputs.env }} IMAGE_TAG=${{ github.event.inputs.image_tag }}
```
dk-singh, line 105: Is `github.event.inputs.env` variable will set as `dev` for manual deployment?

252afh: yeah I think we'll do it that way, afaik dev is the only one we want to be done manually. I've left it configurable for testing in case Lewis needs to use dev, I can just swap to use preprod instead

dk-singh: Make sense.

Makefile
```diff
@@ -193,7 +199,7 @@ tf_apply_universal: ## Apply terraform
 .PHONY: tf_auto_apply
 tf_auto_apply: ## Auto apply terraform
 	make tf_set_workspace && \
-	terraform -chdir=./infrastructure/aws apply -auto-approve -var-file=$(CONFIG_DIR)/${env}-input-params.tfvars ${tf_build_args}
+	terraform -chdir=./infrastructure/aws apply -auto-approve -lock-timeout=300s -var-file=$(CONFIG_DIR)/${env}-input-params.tfvars -var=image_tag=$(IMAGE_TAG) ${tf_build_args} -target="module.django-app" -target="module.core-api" -target="module.worker"
```
lcardno10, line 202: image_tag already gets defined in tf_build_args, so don't think this is necessary

lcardno10 (changes requested):

.github/workflows/manual-release.yml
```diff
@@ -52,44 +60,30 @@ jobs:
     needs:
       - set-vars
       - start-runner
-    uses: i-dot-ai/i-dot-ai-core-github-actions/.github/workflows/apply-terraform.yml@main
+    uses: i-dot-ai/i-dot-ai-core-github-actions/.github/workflows/apply-terraform.yml@353925b7de65e86e6af44027fb2e17e577776319
```
lcardno10, line 63: Change to main

lcardno10 (approved):

PR #515, opened 2024-06-04
Parameterises s3 client for worker and chunker

ghost:
## Context

The s3 client is a global resource for the worker that shouldn't be repeatedly instantiated. This PR adds the client to the other resources treated this way.

Further to this, for evaluation notebooks to use the chunking functions, they need to override and configure the default s3 client. This permits this use of them.

* Parameterises the s3 client in the worker
* Adds `boto3-stubs[essential]` to type it properly

## Guidance to review

* Run `make test-core-api`, `make test-redbox`, `make-test-worker`
* Confirm you're happy with the client added as a global variable
* Adds `worker` to the `core-api` dependencies in the compose so you can bring up the entire backend with `docker compose up core-api`

## Things to check

- [ ] I have run integration tests -- ⚠️ end to end is working, journey isn't

### DIFF
--- a/redbox/parsing/chunkers.py
+++ b/redbox/parsing/chunkers.py
@@ -1,4 +1,5 @@
 from collections.abc import Sequence
+from typing import TYPE_CHECKING
 
 from unstructured.chunking.title import chunk_by_title
 from unstructured.partition.auto import partition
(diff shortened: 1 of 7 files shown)


redbox/parsing/chunkers.py
```diff
@@ -1,16 +1,21 @@
 from collections.abc import Sequence
+from typing import TYPE_CHECKING
```
gecBurton, line 2: TIL

gecBurton (approved):
nice. I prefer this approach, much clearer

PR #518, opened 2024-06-05
Delete expired files with management command

rachaelcodes:
## Context

Now that we have a way to track when files were last used, we can remove them with a management command. There is still some discussion about how this command will be triggered, and this might change over time.

There is a future change to come that will also remove expired Chats and ChatHistory.

Creates a management command that checks for files past their expiry date and removes their chunks from elastic storage, deleted the file contents from S3 and then transitions them to 'deleted' in Django (so we keep a record). 

If there is an error in the elastic or S3 deletion, the file is marked as 'errored' in Django (a new status type). Errors are logged. We need to develop a process of reviewing and handling batches of errors.

### Other changes
* Django views calling the core-api now check for all potential issues with connection, not just HTTP response ones (e.g. connection issues). _We could consider handling S3 errors in future cases_
* 'errored' files are hidden from users as well as 'deleted' ones
* whitespace fixes picked up by the linter in `django_app/tests_playwright/pages.py`

## Guidance to review

Do you agree with the approach? Is there any other error handling needed? What else can we do around files that are partially deleted? 

## Relevant links

## Things to check


brunns (approved):
LGTM.

django_app/redbox_app/redbox_core/management/commands/delete_expired_data.py
```diff
@@ -0,0 +1,53 @@
...
+        for file in File.objects.filter(last_referenced__lt=cutoff_date):
+            logger.debug(
+                "Deleting file object %s, last_referenced %s",
+                file,
+                file.last_referenced,
+            )
+
+            try:
+                core_api.delete_file(file.core_file_uuid, file.user)
+                file.delete_from_s3()
```
gecBurton, line 36: this could be part of the `File.delete` method 🤔 , probably doesnt add much and just over-complicates things?

gecBurton (approved):
super clear, as always!

django_app/redbox_app/redbox_core/management/commands/delete_expired_data.py
rachaelcodes: We currently have a File.delete that actually deletes the file, whereas this just removes the file contents but keeps a record of the File in Django (for auditing processes). I think this was a product decision a few weeks ago.

PR #540, opened 2024-06-07
Feature/semantic routing

andy-symonds:
## Context

This PR adds the semantic routing to `rag chat` using the semantic router library, adding 5 routes, 3 with pre-canned response and 2 placeholder routes for summarisation and extraction.

Additionally, renamed services.py to dependencies.py. We needed to use HuggingFaceEncoder from the semantic-router library - if we use the OpenAIEncoder from the semantic-router library we get nasty errors, which could be a dynamic linking issuing. HuggingFaceEncorder does the job fine, but can be revisited in future.

## Guidance to review

Non-streaming endpoint can be checked using Swagger UI and `test_rag_chat`, however, to test streaming endpoint you would need to go through Django-app. Integration tests added go through streaming endpoint and pass.

The `build_chain` function is used for streaming semantic routing, but does not currently work as-is for the non-streaming endpoint. As a result the non-streaming endpoint has some additional code that can be refactored out in future in favour of using the `build_chain` function. Getting the semantic router library into the codebase and working took priority over and this tech debt should be fixed in due course.

Created a `build_vanilla_chain` function ready to use, but as we are not using vanilla chat just yet, it is not yet used in streaming endpoint.

Once decision to approved, this PR shout be merged to `main` only after PR #539 

## Things to check

- [] I have added any new ENV vars in all deployed environments


gecBurton (approved):
tested in preprod:

* it works very nicely
* i will make some minor formatting changes

core_api/src/semantic_routes.py
```diff
@@ -0,0 +1,107 @@
...
+from redbox.model_db import MODEL_PATH
+
+# === Pre-canned responses for non-LLM routes ===
+INFO_RESPONSE = """
+I am RedBox, an AI focused on helping UK Civil Servants, Political Advisors and
+Ministers triage and summarise information from a wide variety of sources.
+"""
+
+ABILITY_RESPONSE = """
+* I can help you search over selected documents and do Q&A on them.
```
gecBurton, line 14: our markdown formatter in the front end is interpreting:

```
long lines of text that go on and on and on and have new line continuation mark like this \
in them as code and assigns them the <code> tag which isnt what we want
```
image

so i have stripped them out

PR #557, opened 2024-06-11
Show ChatMessage date in admin

rachaelcodes:
## Context
We were asked to show when ChatMessages were updated in admin to make it easier to read. There were a couple of other things I noticed when auditing ChatMessages recently that were annoying and quick to fix. 

- adds 'modified at' to the list view of all the chat histories, so we can see when they were updated
- hides 'selected_files' from the individual chat history view, as it's not used anywhere in the code
- sorts the chat messages within a chat history view by 'modifed_at', so they're seen in the right order
- shows the date each chat message was modified in the chat history view
- makes the 'source_files' for each chat message read only for readability (I looked into adding `filter_horizontal` but it needs more work to make it appear on successive ChatMessageInlines than was worth spending on this task.) 

BEFORE
option 1

AFTER
option 2

## Guidance to review
Have a look at DjangoAdmin locally - is it readable (especially for longer chat conversations)?

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-354

## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/admin.py
+++ b/django_app/redbox_app/redbox_core/admin.py
@@ -18,7 +18,7 @@ class FileResource(admin.ModelAdmin):
 class ChatMessageInline(admin.StackedInline):
     model = models.ChatMessage
     ordering = ("modified_at",)
-    readonly_fields = ["modified_at"]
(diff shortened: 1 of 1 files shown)


brunns (approved):
LGTM.

PR #564, opened 2024-06-12
Feature/capture chunks

gecBurton:
## Context

As a User I want the django app to persist the specific text chunk referenced in RAG so that I can know exactly which part of the document has been referenced.

1. I have renamed the `ChatMessage` `source_files` to `old_source_files`
2. I have created a new model called `TextChunk` to capture the relationship between the `ChatMessage`, `File` and the specific text used in RAG
3. I have added a new m2m field to `ChatMessage` called `source_files` that, like the old field, is a m2m field to `Files` but has a through relationship via `TextChunk`
4. I copy over the data in `old_source_files` to `source_files`
5. I do not delete `old_source_files` in case something goes wrong and or we want to revert this change in production

## Things to check


django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -109,13 +109,13 @@ def save_message(
...
-        source_files: OptFileSeq = None,
+        source_files: MutableSequence[tuple[File, str]] | None = None,
         selected_files: OptFileSeq = None,
     ) -> ChatMessage:
         chat_message = ChatMessage(chat_history=session, text=user_message_text, role=role)
         chat_message.save()
-        if source_files:
-            chat_message.source_files.set(source_files)
+        for file, text in source_files or []:
+            TextChunk.objects.create(chat_message=chat_message, file=file, text=text)
```
gecBurton, line 118: this is the key change

jamesrichards4 (approved):
LGTM but I don't think I can really review the Django side of things in any detail.

On tying this up with other changes, this is still worth doing as retaining Chunks in parts of core-api  is my plan short term. Replacing all the way through the system would be a scary change and I think we can do it incrementally.

PR #575, opened 2024-06-13
Feature/stop streaming

KevinEtchells:
## Context

We need to be able to stop streaming responses for two reasons:
* An accessibility requirement to be able to pause any auto-updating content lasting longer than 5 seconds
* Users may wish to stop the response (e.g. pressed enter too soon, or wanting to ask something slightly different)

The escape key is a standard way to do this, and is recognised as a suitable approach for accessibility.

* Allow streaming to be stopped by pressing the escape key

## Guidance to review

Start a chat, press the escape key

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-368?atlOrigin=eyJpIjoiODM5ZTIyZDFhZDg0NGU1ZjgyNTZlNWE3NjNmOTQ4N2MiLCJwIjoiaiJ9

## Things to think about
* Should we have some text to say "[stream stopped]" or something similar? Or is it clear enough as is?
* We could have a button to stop the stream, but thought it best to wait for the design updates to see if we should add it in then. This PR can be seen as a first step.
* The first commit in this PR added a button (after stopping it) allowing users to resume the stream. I removed this as I felt it was overcomplicating things and we haven't yet identified a user need for this. The code is there, should we wish to revisit this at some point.

### DIFF
--- a/django_app/frontend/js/chats/streaming.js
+++ b/django_app/frontend/js/chats/streaming.js
@@ -44,13 +44,7 @@ customElements.define('sources-list', SourcesList);
 
 class ChatMessage extends HTMLElement {
 
-    constructor() {
-        super();
-        /** Whether the stream has completed - separate to data-status, as it could be paused and completed at the same time */
-        this.complete = false;
-    }
-
-    connectedCallback() {        
+    connectedCallback() {
         this.innerHTML = `
             <div class="iai-chat-message iai-chat-message--${this.dataset.role} govuk-body">
                 <div class="iai-chat-message__role">${this.dataset.role === 'ai' ? 'Redbox' : 'You'}</div>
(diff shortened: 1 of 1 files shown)


django_app/frontend/js/chats/streaming.js
```diff
@@ -76,19 +76,30 @@ class ChatMessage extends HTMLElement {
         let webSocket = new WebSocket(endPoint);
         let streamedContent = '';
         let sources = [];
+
+        // Stop streaming on escape key press
+        this.addEventListener('keydown', (evt) => {
+            if (evt.key === 'Escape' && this.dataset.status === 'streaming') {
+                this.dataset.status = 'stopped';
+                webSocket.close();
```
brunns, line 84: De we need to do anything client-side to deal with this?

brunns: Question - if the stream is escaped from, what if anything should we be saving client-side?

KevinEtchells: It'd be great to save what the user sees, at the point of stopping it. Do you think that's a sensible approach @brunns?

brunns: Might be - probably a question for Charlotte though?

KevinEtchells: I'll speak with Charlotte

KevinEtchells: In case Charlotte doesn't have time to look at it, personally I think the current/default behaviour is okay. If we get user feedback to show otherwise then we can look at it then.

brunns (approved):

PR #574, opened 2024-06-13
Communicate route choice to FE

rachaelcodes:
## Context

Returns the route selected in core_api to the Django FE. Django doesn't do anything with this yet, as there are some design decisions to make.

## Changes proposed in this pull request

- Adds a ChatRouteEnum in `redbox/models` to consolidate route names
- returns 'route' in `ChatResponse` and in the streaming response
- tests the above

## Guidance to review
Have I tested enough? Do I need to return this for all the streaming events?

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-326

## Things to check

- [ ] I have added any new ENV vars in all deployed environments
- [x] I have tested any code added or changed
- [ ] I have run integration tests

PR #588, opened 2024-06-14
Send route name to the front end

rachaelcodes:
## Context
Adds the changes proposed in #574 (which was closed until other things were merged in the mean time).

- Adds a `ChatRoute` class in redbox/models to consolidate route names
- returns 'route' in ChatResponse and in the streaming response
- tests the above

## Guidance to review

Have I tested enough? Am I using the appropriate streaming event

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-326

## Things to check

### DIFF
--- a/core_api/src/routes/chat.py
+++ b/core_api/src/routes/chat.py
@@ -18,7 +18,7 @@ from core_api.src.semantic_routes import (
     INFO_RESPONSE,
     route_layer,
 )
-from redbox.models.chat import ChatRequest, ChatResponse, SourceDocument
+from redbox.models.chat import ChatRequest, ChatResponse, ChatRoute, SourceDocument
 from redbox.storage import ElasticsearchStorageHandler
 
 # === Logging ===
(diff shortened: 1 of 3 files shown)


core_api/src/routes/chat.py
```diff
@@ -45,13 +45,15 @@
...
 
 ROUTE_RESPONSES = {
-    "info": ChatPromptTemplate.from_template(INFO_RESPONSE),
-    "ability": ChatPromptTemplate.from_template(ABILITY_RESPONSE),
-    "coach": ChatPromptTemplate.from_template(COACH_RESPONSE),
-    "gratitude": ChatPromptTemplate.from_template("You're welcome!"),
-    "retrieval": build_retrieval_chain,
-    "summarisation": build_summary_chain,
-    "extract": ChatPromptTemplate.from_template("You asking to extract some information - route not yet implemented"),
+    ChatRoute.info: ChatPromptTemplate.from_template(INFO_RESPONSE),
```
brunns, line 48: Could we make these attributes of the `ChatRoute` objects?

brunns (approved):
LGTM.

gecBurton (commented):
nice to have: extend the integration tests to include this

core_api/src/routes/chat.py
gecBurton: this is likely to be overwritten by https://github.com/i-dot-ai/redbox-copilot/pull/587

rachaelcodes:
> My main concern with the enum approach is that the A/B test stuff proposes lots of ephemeral keyworded routes that will come and go quickly. Would the enum changing quite a lot from deploy to deploy be a problem? I think it isn't, but I thought I'd ask. To the front end it's all just strings right? 

@wpfl-dbt yes it is just strings to the front end, and there is some design work to be done around how to surface the information to users. There is an idea to have some text in the future along the lines of 'this was identified as a summarisation request - was this not what you wanted? try X, Y, Z'. We'll have to make sure the design team are in the loop for the possibility of temporary routes and how to communicate these to users. (Maybe we could have keys with names like 'test-A', 'test-B' or something, or perhaps that is over complicating things). If nothing else, we could have some kind of JS switch block like:

```js
switch (route) {
    case route_options.summary:
        msg = 'X';
        break; 
    case route_options.rag:
        msg = 'Y';
        break;
    default:
        msg = 'Z';
}
```
where `route_options` is provided by an API call surfacing the enum options or sent at the start of the stream or something. The default route would allow us to handle temporary routes.

As for whether to include the Enum - I'm keen on it because we are attempting to type things reasonably strongly and I think there are some routes that we do want to persist.

PR #595, opened 2024-06-17
Create loading ellipsis animation and show until streaming complete

KevinEtchells:
## Context

Feedback has shown we need to show a loading prompt/animation until the stream has finished. It has also suggested an ellipsis animation may be more appropriate than a spinner

* Swap out the spinner for an ellipsis
* Show it until streaming is complete

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-382

## Questions

* Is this better than the spinner?
* Should we show this instead of the spinner, or as well as it?
* Is the styling okay?
* Any other thoughts?


brunns (approved):
So, it's definitely working great. (Just remember to `poetry run python manage.py collectstatic`, and refresh your browser cache.) Looks good to me, too, but I have notoriously bad judgement when it comes to design, so make of that what you will.

django_app/frontend/style.scss
```diff
@@ -183,6 +183,39 @@ body {
   text-align: left;
 }
 
+.rb-loading-ellipsis {
+  font-style: italic;
+}
```
rachaelcodes, line 188: 👍

PR #596, opened 2024-06-17
Refactored the chat app to be fastapi dependency based

jamesrichards4:
## Context

Refactoring the chat app to pull the chains (retrieval and summary) as FastAPI dependencies. This should allow overriding and testing more easily and creates a clean way to add new routes and chains. Lots more to do on this, particularly around the testing but this is the blocking part due to the chat module changes.

Made the path functions in the chat app pull the routes dictionary as a fastapi dependency, the dependency chain goes all the way back to the retriever and llms now which means they can be overridden in testing.
Mocked the llm out in tests in test_chat which is the start of enabling testing new chains without too many dependencies (we may be able to parameterise out the retriever too to allow proper testing?)

## Things to check

### DIFF
--- a/core_api/src/dependencies.py
+++ b/core_api/src/dependencies.py
@@ -148,8 +148,3 @@ def get_llm(env: Annotated[Settings, Depends(get_env)]) -> ChatLiteLLM:
         log.exception(msg)
         raise ValueError(msg)
     return llm
-
-
-# Hack before refactoring to dependency managed chains
-es_retriever = get_es_retriever(get_env(), get_elasticsearch_client(get_env()), get_embedding_model(get_env()))
-llm = get_llm(get_env())
--- a/core_api/src/semantic_routes.py
+++ b/core_api/src/semantic_routes.py
@@ -1,7 +1,12 @@
+from typing import Annotated
+
+from fastapi import Depends
+from langchain_core.runnables import Runnable
 from semantic_router import Route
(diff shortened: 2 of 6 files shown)


jamesrichards4:
Just moved the prompts to be fixed for build_chains to avoid some weird impact on the api listing thinking they could be passed in. They should be moved to env vars anyway under another ticket

core_api/src/dependencies.py
```diff
@@ -148,8 +148,3 @@ def get_llm(env: Annotated[Settings, Depends(get_env)]) -> ChatLiteLLM:
         log.exception(msg)
         raise ValueError(msg)
     return llm
-
-
-# Hack before refactoring to dependency managed chains
```
gecBurton, line 153: :+1:

core_api/src/semantic_routes.py
```diff
@@ -101,7 +106,30 @@
 )
 
 
-routes = [info, ability, coach, gratitude, summarisation, extract]
+def get_semantic_routes():
+    return [info, ability, coach, gratitude, summarisation]
+
+
+def get_semantic_routing_encoder():
+    return HuggingFaceEncoder(name="sentence-transformers/paraphrase-albert-small-v2", cache_dir=MODEL_PATH)
```
gecBurton, line 114: note to self, this should come from Settings.emebedding_model (not urgent though)

gecBurton (approved):

PR #598, opened 2024-06-17
Patches for streaming

jamesrichards4:
## Context

Integration tests are failing on main

The handling of page numbers in ChatResponse now accounts for all possibilities. This fix is ugly, there's a better fix in there somewhere but it needs thinking about.
With the move to LCEL runnables for chains we've lost access to some of the streaming events. This breaks the fixed prompt based responses so this has been patched. This fix matches the existing behaviour ( I think! ) of returning text emitted by the prompt and the LLM but this means the retrieval pipeline will return both the prompt and the LLM response which feels wrong? The test passes but I'm not sure about this behaviour

## Guidance to review

The websocket streaming chat path function seems to be doing what was done before but if somebody can shed any light on how the frontend is going to handle receiving event messages for the prompt and LLM that'd be good.

## Things to check

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,7 +8,8 @@
     },
     "python.analysis.autoImportCompletions": true,
     "python.testing.pytestArgs": [
-        "core_api"
+        "core_api",
+        "tests"
     ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true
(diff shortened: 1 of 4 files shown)


lmwilkigov (approved):

gecBurton (approved):
thx, LGTM for now will rethink this asap

PR #607, opened 2024-06-18
bugfix/remove-prompt-from-output

gecBurton:
## Context

As a User I do not want the prompt returned in the output

## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/9563515191


jamesrichards4:
I thought removing this part would fail on the fixed responses? Looks like the tests pass though so we're still getting a gratitude response in integration tests?

gecBurton:
> I thought removing this part would fail on the fixed responses? Looks like the tests pass though so we're still getting a gratitude response in integration tests?

ah - good point. ill find a solution

PR #611, opened 2024-06-18
REDBOX-234 show file expiry

brunns:
## Context

User need to see when their files will expire.

Show user when their file will expire - including taking account of the fact they may already have expired. 

Also fix the bug that streamed chat wasn't updating the file's last referenced date.

Screenshot 2024-06-18 at 15 13 14

## Things to check


gecBurton (approved):
LGTM

PR #612, opened 2024-06-18
bugfix/add-routes-to-output

gecBurton:

1. added route to the response payload
2. use the route to ensure that we do not send back the rag/summary prompt (this is very hacky but will do for now)
3. added hash to DocumentSource so it can be de-duplicated

## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/9566739959

### DIFF
--- a/core_api/src/routes/chat.py
+++ b/core_api/src/routes/chat.py
@@ -39,7 +39,7 @@ chat_app = FastAPI(
 
 async def semantic_router_to_chain(
     chat_request: ChatRequest, user_uuid: UUID, routable_chains: dict[str, Runnable], route_layer: RouteLayer
-) -> tuple[Runnable, ChainInput]:
+) -> tuple[Runnable, ChainInput, str]:
     question = chat_request.message_history[-1].text
     route = route_layer(question)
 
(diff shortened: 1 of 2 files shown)


core_api/src/routes/chat.py
```diff
@@ -79,36 +79,41 @@ async def rag_chat_streamed(
...
+    selected_chain, params, route = await semantic_router_to_chain(
+        chat_request, user_uuid, routable_chains, route_layer
+    )
 
     async for event in selected_chain.astream_events(params.model_dump(), version="v1"):
         kind = event["event"]
+        log.info("%s %s %s", kind, route, event["data"])
         if kind == "on_chat_model_stream":
-            await websocket.send_json({"resource_type": "text", "data": event["data"]["chunk"].content})
+            await websocket.send_json({"resource_type": "text", "data": event["data"]["chunk"].content, "route": route})
```
brunns, line 90: Does every bit of text need the route? It's one route for the entire response, right?

gecBurton: fair point. killing this PR in anycase

PR #613, opened 2024-06-18
Bugfix/duplicated prompt in response

jamesrichards4:
## Context
We create some of our dependencies every request. This moves the semantic routes resources to be singletons but still allows overriding in tests
We also return the prompt in our streamed responses

Use atsream chains in chat path function to make consuming llm responses simpler, we don't return the prompt and response now
Made the semantic route dependencies singletons

## Guidance to review

these could be two PRs if we want? There's a branch with just the semantic routes dependency work

## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/9567410892

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,7 +12,7 @@
             "args": [
                 "core_api.src.app:app",
                 "--reload",
-                "--port", "8080"
+                "--port", "5002"
             ],
             "jinja": true,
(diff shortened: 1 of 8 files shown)


gecBurton (approved):
LGTM pending IRL tests on dev and integrations tests

PR #635, opened 2024-06-24
Added vanilla route

lmwilkigov:
## Context

Redbox currently doesn't handle off topic conversations well, where normal chat services do.

I've added a vanilla route which aims to act as broad collection of phrases that aren't directly information seeking. This then routes to a new chain that allows a minimally prompted LLM to respond directly with the full chat history.

## Guidance to review

Example below on asking Redbox to do something odd.

Screenshot 2024-06-24 at 09 42 02

To review, load up UX and try a few requests into the model, and some RAG ones to check intersection of routes.

The routing will require refinement with more utterances!

## Things to check

### DIFF
--- a/core_api/src/build_chains.py
+++ b/core_api/src/build_chains.py
@@ -25,30 +25,27 @@ log = logging.getLogger()
 
 
 def build_vanilla_chain(
-    chat_request: ChatRequest,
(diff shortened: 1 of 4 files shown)


gecBurton (approved):
nicely done

PR #642, opened 2024-06-24
Journey tests - ensure correct route is use, as selected by keyword

brunns:
## Things to check


tests/test_journey.py
```diff
@@ -67,7 +67,7 @@ def test_user_journey(page: Page, email_address: str):
     logger.debug("page: %s", chats_page)
     logger.info("all_messages: %s", chats_page.get_all_messages_once_streaming_has_completed())
     latest_chat_response = chats_page.wait_for_latest_message()
-    assert "architecture" in latest_chat_response.text
```
gecBurton, line 70: :+1: this is a leading cause of false negative tests i think?

gecBurton (approved):
LGTM

tests/test_journey.py
brunns: Yup, and it's not the right place to be testing response quality anyway I think.

PR #643, opened 2024-06-24
Redbox 411 documents refactor

jamesrichards4:
## Context

To allow removing some custom code and make greater use of langchain components and features we need to make Documents not Chunks the main data representation of text chunks. To separate this into stages this part focusses on the worker and loads ingested files into a new Document structure using langchain classes; this can still be accessed using our current retriever with some additional filters and mapping functionality.

Removal of chunks will follow once we've pulled this through to the API layer and no longer need chunks.

Rewrite of worker to be LCEL based. This makes greater use of Langchain and should enable us to create configurable ingest chains to test different values/features etc

## Guidance to review

I haven't touched the File index. This leaves the current status methods etc in place. To remove the storage handler code entirely we'd need to remove this or replace it with another Document index (hierarchical Documents style). This seems ok for this change but thoughts welcome

This PR is an incremental move to a more langchainy way of doing things but it is worse in some places as we transition. This is to allow users to use pre-move files, and to explore doing this kind of backwards compatible ingest change. We can do away with it if we're ok with a clean break. The additional work on the refactor will follow shortly and the removal of legacy handling in the retriever can go once old files are expired 

## Things to check


jamesrichards4:
Moved to feature branch for PR checks

PR #644, opened 2024-06-24
Feature/redbox 411 documents refactor

jamesrichards4:
## Context

To start using Documents rather than custom chunks. The intention is to replace all use of chunks in the 'AI' code in a second PR following this one

Update the retriever and storagehandler to be able to use chunk and document style stored chunks. This means documents ingested either way are fine.

Worker rewritten to be LCEL based. This should allow moving to Azure embeddings more easily and is in line with the rest of the codebase

## Guidance to review

The storage handler currently handles File objects and can track status. To drop the storage handler chunk access we need to find a way to update the API about status. Moving the storage handler to being a file handler only and having the worker update the status makes this quite clean but other ideas welcome on how we make the most of these changes

There are several places logic is duplicated, this has been left as it will be replaced by dropping the storagehandler version of this logic once we can drop chunk storage (and use documents only). This will come when they expire in all environments.

This will be followed by several PRs:
* One to complete the use of documents and make the retriever use documents from Elastic and pass them through to SourceDocuments to the frontend, this will remove all references to Chunks. The storagehandler logic will also be updated here.
* Move to Azure Embeddings in ingest and API to remove the need for torch, sentencetransformers etc.
* Dropping duplicate logic for fields on stored chunks and use just the document structure in the 'AI' side code. These can be mapped to SourceDocuments and other API objects in defined runnables

## Things to check


gecBurton:
> To drop the storage handler chunk access we need to find a way to update the API about status. 

It might help to know that we might not need the same level of granularity that we currently have. Users have told us that they arent interested in the uploades/chunkinging/embedding/complete status but only want to know if the file is ready/not-ready cc @KevinEtchells

redbox/storage/elasticsearch.py
```diff
@@ -210,3 +232,21 @@ def get_file_status(self, file_uuid: UUID, user_uuid: UUID) -> FileStatus:
             chunk_statuses=chunk_statuses,
             processing_status=ProcessingStatusEnum.complete if is_complete else ProcessingStatusEnum.embedding,
         )
+
+
+def hit_to_chunk(hit: Dict[str, Any]) -> Chunk:
```
gecBurton, line 237: nice. this "inclusive" approach looks like the best way to start this major refactor.

worker/src/loader/file_loader.py
```diff
@@ -0,0 +1,63 @@
...
+            file: The RedBox File to load
+        """
+        self.file = file
+        self.s3_client = s3_client
+        self.env = env
+
+    def lazy_load(self) -> Iterator[Document]:  # <-- Does not take any arguments
+        """A lazy loader that reads a file line by line.
+
+        When you're implementing lazy load methods, you should use a generator
```
gecBurton, line 35: but you dont?

gecBurton (approved):
I have understood the approach as:
1. all new text-chunks will be encoded as `Document`s
2. on reading text-chunks they will be cast to `Chunk`s regardless of how they we encoded

In which case it all looks sensible to me.


I presume the next stage is to change step (2) to return `Document`s? Are we thinking that:
1. we should maintain backwards compatibility for text-chunks encoded as `Chunk`s?
2. we should write one off migration script
4. we should accept a breaking change for July

worker/src/loader/file_loader.py
jamesrichards4: Good point :)  I used the DocumentLoader structure but actually the next step in the runnable wants a list so it can batch things. Cleanest way would be to collect the generator in a runnable between the two. I'll sort it in the next PR for this?

redbox/models/settings.py
```diff
@@ -125,8 +125,12 @@ class Settings(BaseSettings):
 
     openai_api_version: str = "2023-12-01-preview"
     azure_openai_model: str = "azure/gpt-35-turbo-16k"
+    azure_embedding_model: str = "text-embedding-ada-002"
```
lmwilkigov, line 128: Just to double check, we only have 350K TPM approved for ada, but we do have 700K TPM for text-embedding-3-large which is more performant. Do we want to swap out now?

lmwilkigov: Clarified that this is happening in later PR

worker/src/loader/file_loader.py
jamesrichards4: Fixed here because mypy was whinging

jamesrichards4:
I'll handle the conflicts after https://github.com/i-dot-ai/redbox-copilot/pull/630 goes in. We should merge that first

lmwilkigov (approved):

PR #638, opened 2024-06-24
REDBOX-405 create changelog.

brunns:
## Things to check

- [ ] I have added any new ENV vars in all deployed environments
- [x] I have tested any code added or changed
- [x] I have run integration tests

PR #645, opened 2024-06-24
Feature/updated docs and versions

lmwilkigov:
## Context

Versions weren't consistent and docs were broken

I've propagated updated version through pyproject.toml, library and core api. I've also spruced up the docs with a gif of the UI in action.

## Guidance to review

Check out the status endpoint in core api and build the docs

## Things to check

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ target-version = 'py311'
 
 [tool.poetry]
 name = "redbox"
-version = "0.0.3"
+version = "0.3.0"
 description = "RAG application that uses GenAI to chat with and summarise civil service documents"
 authors = ["i.AI <i-dot-ai-enquiries@cabinetoffice.gov.uk>"]
 license = "MIT"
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ target-version = 'py311'
 
 [tool.poetry]
 name = "redbox"
-version = "1.0.0"
+version = "0.0.3"
```
brunns, line 7: Aren't we on `0.3.0`?

```suggestion
version = "0.3.0"
```

redbox/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Redbox is a Python library for working with the Redbox API, data and services."""
 
-__version__ = "0.9.0"
+__version__ = "0.0.3"
```
brunns, line 3:
```suggestion
__version__ = "0.3.0"
```

gecBurton (approved):
LGTM

we should think about a GH action to ensure that the version number is correct on release

brunns:
> LGTM
> 
> we should think about a GH action to ensure that the version number is correct on release

And also generate the changelog, assuming we think #638 is the right approach.

brunns (approved):
LGTM.

PR #621, opened 2024-06-20
Send route name to the front end

rachaelcodes:
## Context

Adds the changes proposed in https://github.com/i-dot-ai/redbox-copilot/pull/574 and https://github.com/i-dot-ai/redbox-copilot/pull/588 (which were both closed until other things were merged in the meantime).

## Changes proposed in this pull request

- Adds a ChatRoute class in redbox/models to consolidate route names
- returns 'route_name' in ChatResponse and in the streaming response
- tests the above
- TODO: add to e2e test

## Guidance to review

Have I tested enough? Is the ChatRoute Enum actually useful?

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-326

## Things to check

- [ ] I have added any new ENV vars in all deployed environments
- [ ] I have tested any code added or changed
- [ ] I have run integration tests

PR #648, opened 2024-06-25
Add route_name to e2e tests

rachaelcodes:
## Context
Updates the e2e tests as I didn't have time when writing #621

Added `route_name` checks to e2e tests.

Have moved the old `test_post_rag_stuff` test so it's sat with the other non-streaming tests and renamed it to `test_post_rag_summarisation` to reflect the route.

## Guidance to review

Is it worth adding anything else to these? Do you agree with the test name change?

## Things to check


tests/test_e2e.py
```diff
@@ -150,6 +156,30 @@ def test_post_rag_fail(self, user_uuid):
         }
         assert TestEndToEnd.file_uuids[user_uuid] not in source_document_file_uuids
 
+        TestEndToEnd.route_name = rag_response.json()["route_name"]
+        assert TestEndToEnd.route_name == ChatRoute.retrieval
+
+    @pytest.mark.parametrize("user_uuid", USER_UUIDS)
+    def test_post_rag_summarisation(self, user_uuid):
```
rachaelcodes, line 163: this is the test that moved from the end of the file

gecBurton (approved):
LGTM

PR #656, opened 2024-06-25
Add parallel processing to `map` step of map reduce summarisation

andy-symonds:
## Context

Before this change, the map step of map reduce summarisation, used for large document summarisation, was not happing in parallel. The code needed to be updated to allow the parallele summarisation of large documents into smaller summaries using LCEL `batch`.

`batch` only takes a list as input, so I needed to split the question prompt into `map_question_prompt`, which is populated with the question and chat_history before being passed to the `map_operation` and then the `map_document_prompt` which is populated when running `batch`.

- Updated `build_map_reduce_summary_chain` to allow for `batch` processing
- Made `Chunk` model iterable
- Renamed `max_tokens` in env var AI Settings to `summarisation_chunk_max_tokens` to better describe what it does. `max_tokens` is already an overloaded term.

## Guidance to review

Check you are happy with how I have structured `build_map_reduce_summary_chain`.

## Things to check

### DIFF
--- a/core_api/src/build_chains.py
+++ b/core_api/src/build_chains.py
@@ -34,9 +34,7 @@ def build_vanilla_chain(
     env: Annotated[Settings, Depends(dependencies.get_env)],
 ) -> Runnable:
     return (
-        make_chat_prompt_from_messages_runnable(
-            env.ai.vanilla_system_prompt, env.ai.vanilla_question_prompt
-        )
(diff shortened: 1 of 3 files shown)


jamesrichards4 (approved):
Looks good. Probably worth making max_concurrency a setting so we can switch it out if we get issues though

PR #674, opened 2024-06-26
Add an environment variable to set the context window size

ghost:
## Context

We're getting errors because [the context window size is being exceeded](https://technologyprogramme.atlassian.net/jira/software/projects/REDBOX/boards/409?selectedIssue=REDBOX-417).

* Adds an environment variable to declare the maximum
* Adds a tiktoken dependency to count tokens
* Uses it in the chat history, and nowhere else for now

## Guidance to review

* Check you're happy with how I've implemented this
* Run the django app and check it works

## Things to check

- [x] [I have run integration tests](https://github.com/i-dot-ai/redbox-copilot/actions/runs/9683170696)

### DIFF
--- a/core_api/tests/test_runnables.py
+++ b/core_api/tests/test_runnables.py
@@ -1,4 +1,3 @@
-
 import pytest
 
 from core_api.src.app import env


brunns:
Do we need to add this to any of the `.env` files?

brunns:
If it picks up sensible defaults elsewhere, that's fine I think. But if we need them set, we should set them in `.env.integration` at least.

Thinking about it, if the tests are passing, we clearly don't need them explicitly set in `.env` so it's fine, right?

core_api/src/runnables.py
```diff
@@ -10,24 +10,47 @@
...
-def make_chat_prompt_from_messages_runnable(system_prompt: str, question_prompt: str):
+def make_chat_prompt_from_messages_runnable(
+    system_prompt: str,
+    question_prompt: str,
+    context_window_size: int,
+    tokeniser: Encoding,
+):
     system_prompt_message = [("system", system_prompt)]
+    prompts_budget = len(tokeniser.encode(system_prompt)) - len(tokeniser.encode(question_prompt))
+    token_budget = context_window_size - prompts_budget
```
jamesrichards4, line 29: This needs to take into account the output max tokens (llm.max_tokens)

core_api/src/runnables.py
```diff
@@ -10,24 +10,47 @@
...
     def chat_prompt_from_messages(input_dict: dict):
         """
         Create a ChatPrompTemplate as part of a chain using 'chat_history'.
         Returns the PromptValue using values in the input_dict
         """
+        context_budget = token_budget - len(tokeniser.encode(input_dict["question"]))
+
+        if context_budget <= 0:
+            message = "Question length exceeds context window."
+            raise AIError(message)
```
jamesrichards4, line 41: This gives a 500 in the API response? What about Django?

brunns: Sorry, not sure I understand the Django question.

jamesrichards4 (approved):

PR #679, opened 2024-06-27
Auto-expand message input to match content

KevinEtchells:
## Context

Following the design system updates, the message input is only 1 row high. That could create problems for users when entering long messages.

* Make the height automatically update to fit the content
* Set a minimum height of 1 row
* Set a maximum height of 5 rows
* Prevent horizontal resizing

## Things to check

* That it expands and contracts automatically
* That it doesn't expand past 5 rows


gecBurton (approved):
works for me!

PR #680, opened 2024-06-27
Switched to Azure embeddings in semantic router, api and worker

jamesrichards4:
NB: This will BREAK users ability to work with files already ingested. Once deployed to dev it will be difficult/impossible to promote any changes without handling this breaking change

On deploy we should update the env var EMBEDDING_DOCUMENT_FIELD_NAME to store the new embeddings in a field called something like 'azure_embedding'. This should avoid issues with the field mapping as we move to different size embeddings

## Context

We're using local embeddings for semantic routing and RAG. We want to use a remote embedding service to speed up ingest and remove the complicated dependencies from worker and api images

Switch worker to use Langchain AzureOpenAIEmbeddings in the ElasticsearchStore
Switch core-api embeddings dependency to be AzureOpenAIEmbeddings
Switch semantic router EmbeddingEncoder to be AzureOpenAIEncoder

Some refactoring of embeddings in tests has also been done. All tests now use FakeEmbeddings at a fixed dimension, equal to the model set in example env. The semantic routes encoder uses TfidfEncoder in tests which is fragile on the utterances.

## Guidance to review

Are we happy with the increased size of these embeddings 756 -> 3072 this might make moving them about slower?
Are we happy with the model choice and the increased costs? We should be able to reduce the size of workers to compensate.

Liam has agreed the above two changes.

Have I caught all the additional dependencies we no longer need? sentence-transformers and torch are out, not sure there's anything else? Could be worth building images from scratch now to get clean ones with no cached layers?

## Things to check


core_api/src/dependencies.py
```diff
@@ -35,9 +34,14 @@ def get_elasticsearch_client(env: Annotated[Settings, Depends(get_env)]) -> Elas
 
 @lru_cache(1)
 def get_embedding_model(env: Annotated[Settings, Depends(get_env)]) -> Embeddings:
-    embedding_model = SentenceTransformerEmbeddings(model_name=env.embedding_model, cache_folder=MODEL_PATH)
-    log.info("Loaded embedding model from environment: %s", env.embedding_model)
-    return embedding_model
+    return AzureOpenAIEmbeddings(
+        azure_endpoint=env.azure_openai_endpoint,
+        openai_api_version="2023-05-15",
```
lmwilkigov, line 39: Should this be using env.openai_api_version instead of hardcoded? It mismatches our current default version, so we might want to consider updating to the version you note here.

core_api/src/semantic_routes.py
```diff
@@ -147,11 +147,11 @@ def get_semantic_routes():
...
 
 def get_semantic_routing_encoder():
+    env = get_env()
     global __semantic_routing_encoder  # noqa: PLW0603
     if not __semantic_routing_encoder:
-        __semantic_routing_encoder = HuggingFaceEncoder(
-            name="sentence-transformers/paraphrase-albert-small-v2",
-            cache_dir=MODEL_PATH,
+        __semantic_routing_encoder = AzureOpenAIEncoder(
+            azure_endpoint=env.azure_openai_endpoint, api_version="2023-05-15", model=env.azure_embedding_model
```
lmwilkigov, line 154: Similar Q on getting the API version from ENV

core_api/tests/conftest.py
```diff
@@ -95,8 +94,8 @@ def stored_file_1(elasticsearch_storage_handler, file) -> File:
 
 
 @pytest.fixture()
-def embedding_model_dim(embedding_model) -> int:
-    return len(embedding_model.embed_query("foo"))
+def embedding_model_dim() -> int:
+    return 1536 #text-embedding-3-small default dimension
```
lmwilkigov: For text-embedding-3-large the dimension size is 3072

https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#embeddings-models

lmwilkigov (changes requested):

core_api/tests/conftest.py
jamesrichards4: do we want to use text-embedding-3-large in development? It would be cleaner but I'm thinking about costs? We can use ada-002 in our region is it worth doing that?

jamesrichards4: We're all good for cost concerns for now, going to use 3-large everywhere

lmwilkigov:
https://github.com/i-dot-ai/redbox-copilot/blob/be75e55fc6b210f629975d89effbff1efa42158c/pyproject.toml#L96 This leftover in the pyproject.toml is causing a build failure

lmwilkigov (approved):

core_api/tests/routes/test_chat.py
```diff
@@ -70,13 +71,21 @@ def mock_all_chunks_retriever(alice):
     return RunnableLambda(lambda _: docs)
 
 
-@pytest.fixture()
+def mock_semantic_route_encoder():
+    encoder = TfidfEncoder()
```
gecBurton, line 75: 👍

core_api/src/semantic_routes.py
```diff
@@ -147,20 +149,22 @@ def get_semantic_routes():
     return (info, ability, coach, gratitude, summarisation, vanilla)
 
 
-def get_semantic_routing_encoder():
-    global __semantic_routing_encoder  # noqa: PLW0603
```
gecBurton, line 151: pleased to see this necessary evil go :)

pyproject.toml
```diff
@@ -18,9 +18,7 @@ langchain = "^0.2.6"
 tiktoken = "^0.7.0"
 boto3 = "^1.34.136"
 pydantic-settings = "^2.3.4"
-sentence-transformers = "<3.0"
-unstructured = {version = "<0.14.9", extras = ["all-docs"]}
```
gecBurton, line 22: my bad, i should have spotted this, same for torch below

gecBurton: > my bad, i should have spotted this, same for torch below

infrastructure/aws/variables.tf
```diff
@@ -262,4 +262,28 @@ variable "llm_max_tokens" {
...
+}
+
+variable "embedding_max_retries" {
+  type          = string
+  default       = 10
+  description   = "Number of retries to external embedding services (rate limiting)"
+}
+
+variable "embedding_retry_min_seconds" {
+  type          = int
```
gecBurton, line 280:
```suggestion
  type          = number
```

gecBurton:
building https://github.com/i-dot-ai/redbox-copilot/actions/runs/9807182035

gecBurton:
integration teats https://github.com/i-dot-ai/redbox-copilot/actions/runs/9807188354

gecBurton:
integration tests https://github.com/i-dot-ai/redbox/actions/runs/9809337402

PR #681, opened 2024-06-27
Add logic within summarisation route to select appropriate stuff or map reduce method

andy-symonds:
## Context

When a user wants to summarise a document, they are not concerned with the method that is used to do the summarisation. Under-the-hood, however, when going down the summarisation route `core-api` needs to select the appropriate method, `stuff` for basic summarisation or `map reduce` for large document summarisation.

Reduced summarisation runnable to a single `build_summary_chain` that uses logic within the langchain runnable to decide which summarisation method to use, based on the size of the document, set by `summarisation_chunk_max_tokens`. 

`core-api` now passes back to the frontend which summarisation method was used.

## Things to check


redbox/models/chat.py
```diff
@@ -69,6 +70,8 @@ class ChatRoute(StrEnum):
     gratitude = "gratitude"
     search = "search"
     summarise = "summarise"
+    stuff_summarise = "basic summarisation"
```
jamesrichards4, line 73: This is a tiny thing but if we want to use these chat route names as keywords they'll need to have no spaces. Is it worth maintaining snake_case so we always have that option?

lmwilkigov: Yes, I think our current regex would only match on basic in this case.

tests/test_journey.py
```diff
@@ -89,7 +89,6 @@ def test_user_journey(page: Page, email_address: str):
         chats_page = chats_page.send()
         latest_chat_response = chats_page.wait_for_latest_message()
         assert latest_chat_response.text
-        assert latest_chat_response.route == route
```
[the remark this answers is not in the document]
jamesrichards4: we have to as summarisation now returns the route name for the delegated route which is unknown from the user end. It's hard to test without exposing a lot of info as the name that comes back is based on internals

PR #684, opened 2024-06-28
Don't attempt to upload large files in the journey test.

brunns:
Also, wait for footer to load before doing a11y checks, so see if this fixes the transient failures.

## Things to check


gecBurton (approved):
lgtm

PR #685, opened 2024-06-28
Added Azure env vars to worker secrets

jamesrichards4:
## Context

We're using Azure in the worker now so we need to set Azure API key and endpoint values.

This PR is most me checking my understanding of how the env var/terraform side of things works

Set the Azure OpenAI API key and endpoint to the same vars as used in core api

## Guidance to review

The worker tests also need these vars set. My feeling is this shouldn't be necessary and I should rework the tests to avoid this?

## Things to check


jamesrichards4:
Not needed. These changes are in the redbox-430 feature PR now

PR #691, opened 2024-06-28
[BUGFIX] All documents returned for the summariser

ghost:
## Context

The all document retriever wasn't bringing back all documents -- it was limited to 10. Fixed with a new class, which in turn throw up some more problems that needed fixing.

* Adds `AllElasticsearchRetriever` to return all chunks in order
    * Removed embeddings from the returned object
    * Implemented `scan()` to get all results
* Enforces the shape of returned documents at the retrieval class level, and simplifies the returned structure
    * `Chunk` and `Document` both handled
* Refactors `core_api.src.retriever` to be simpler for the new classes and functions\
* Refactors API to use these changes

## Guidance to review

* Look through `core_api.src.retriever` most closely -- this is where the action is
* Ensure it all runs

## Things to check

- [x] [I have run integration tests](https://github.com/i-dot-ai/redbox-copilot/actions/runs/9715657654)

### DIFF
--- a/core_api/tests/routes/test_chat.py
+++ b/core_api/tests/routes/test_chat.py
@@ -52,11 +52,7 @@ def mock_parameterised_retriever(alice):
     docs = [
         Document(
             page_content="some text that doesn't actually matter " * 10,
-            metadata={
-                "_source": {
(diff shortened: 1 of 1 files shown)


jamesrichards4 (approved):
Looks great, this is what my refactor should have been in the first place

PR #695, opened 2024-07-01
Remove S3 dependency from chunking process

wpfl-dbt:
## Context

In order to regularly test the AI in the system, we need:

* To import and use the chunking functions at the highest level we can
* Avoid mocking/simulating unnecessary dependencies

For these reasons this PR removes S3 from the chunking process, instead passing in a bytestream. In production this just means downloading from S3 higher up the process. For testing, this means we can use the chunking code directly without needing MinIO/S3 in the picture.

* Refactors chunking so S3 passes the opened file into it
* Removes dead `redbox.parsing` code and tests
* Updates `quickupload.ipynb` so it better reflects code in the worker

## Guidance to review

* Check you're happy with the refactor
* Confirm all unit tests pass

## Things to check

- [x] [I have run integration tests](https://github.com/i-dot-ai/redbox-copilot/actions/runs/9740019023)


notebooks/quickupload.ipynb
```diff
@@ -114,7 +123,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "USER_UUID = UUID(\"5c37bf4c-002c-458d-9e68-03042f76a5b1\")"
+    "USER_UUID = RB_APP.User.objects.all()[0].id # or read it and hardcode\n",
```
gecBurton, line 126:
```suggestion
    "USER_UUID = RB_APP.User.objects.first().id # or read it and hardcode\n",
```

worker/src/app.py
```diff
@@ -66,7 +67,10 @@ async def lifespan(context: ContextRepo):
 def document_loader(s3_client: S3Client, env: Settings):
     @chain
     def wrapped(file: File):
-        return UnstructuredDocumentLoader(file, s3_client, env).lazy_load()
+        file_raw = BytesIO()
+        s3_client.download_fileobj(Bucket=file.bucket, Key=file.key, Fileobj=file_raw)
```
jamesrichards4, line 71: can we pass the 'Body' field of get_object to the documentloader to avoid this download? It's also a byte stream so that should work?

jamesrichards4 (commented):
Love the idea, just one comment on whether we can avoid downloading the file. We should think about cleaning them up if we've pulled them to the worker

worker/src/app.py
jamesrichards4: ah ok yeah in which case let's pull it. We should think about cleaning these files up at some point but maybe don't need to worry in this release. Restaring/scaling the workers will clear out all files anyway so should be ok for a bit

jamesrichards4 (approved):

PR #696, opened 2024-07-01
Feature/redbox 367 show citation text

brunns:
## Context

Allow the user to see some detail of the citations.

Screenshot 2024-07-01 at 13 12 38

Screenshot 2024-07-01 at 13 12 48

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-367

## Things to check


django_app/redbox_app/redbox_core/models.py
```diff
@@ -209,6 +209,9 @@ def expires_at(self) -> datetime:
     def expires(self) -> timedelta:
         return self.expires_at - datetime.now(tz=UTC)
 
+    def __lt__(self, other):
+        return self.id < other.id
```
rachaelcodes, line 213: Would `created_at` be a better metric to sort by? Or does the order not matter?

brunns: Right now, it doesn't matter - we only need a total ordering so that we can group citations by file. If one day we need some specific order, we can revisit.

rachaelcodes (commented):
One problem of this approach is that the order of the source responses is lost on the citations page. 

e.g. on the chat view - which has the most relevant source file listed first: 
chat_sources

and in the citations page view - which loses this ordering
sources_with_citations

Given that this is the 'quick' initial round of showing citations, I would be happy with this discrepancy if Charlotte and Gabe are.

django_app/redbox_app/redbox_core/models.py
rachaelcodes: I've put in another comment about the grouping by file - can we preserve the order of the 'most relevant file'? Or is that for a later PR?

brunns: Right now we don't store the citation order at all AFAICT. We might need to revisit that.

rachaelcodes (approved):
A good quick pass at this, which I think will add a lot of user value. Approved pending approval from Gabe/Charlotte re the discrepancy in File order.

PR #708, opened 2024-07-02
feature/package-rebox-core

gecBurton:
## Context

As an Engineer I want to be able to package the redbox-core library so that I can install it from applications such as:
* core-api
* worker
* other

* `redbox/` has been renames `redbox-core/`
* `redbox-core` has its own `pyproject.toml`
* all src code has been moved into a subdirectory called `redbox` as per poetry spec 
* various other minor adjustments

## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox-copilot/actions/runs/9757962922


pyproject.toml
```diff
@@ -12,6 +12,7 @@ readme = "README.md"
 
 [tool.poetry.dependencies]
 python = ">=3.11,<3.12"
+redbox = {path="redbox-core", develop=true}
```
gecBurton, line 15: this is the key change

.gitignore
```diff
@@ -1,5 +1,6 @@
 **/*.db
 .idea/
+redbox-core/models/
```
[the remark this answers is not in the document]
gecBurton: these are the models cached my huggingface

pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 120
 target-version = 'py311'
 
 [tool.poetry]
-name = "redbox-copilot"
+name = "redbox-root"
```
gecBurton, line 6: cant be called `redbox` otherwise we get:

```
The Poetry configuration is invalid:
  - Project name (redbox) is same as one of its dependencies
```

PR #709, opened 2024-07-02
Feature/package worker

gecBurton:
## Context

As an engineer I want all applications (in this case the `worker`) to have their own `pyproject.toml` so that it isn clear what python-env is being deployed in any given Docker container

1. `worker` has its own `pyproject.toml` which has `redbox` as a dependancy
2. `worker/loader` has been moved to `redbox-core`:
     * this reflect the fact that another app might want to load documents
     * it means that there is now many less dependencies in the `worker`
3. `worker` has its own GH action test job
4.  unused conftest code has been removed

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox/actions/runs/9838087187


.github/workflows/python-app.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         cp .env.test .env
         docker compose up -d --wait elasticsearch
 
-        poetry install --no-root --no-ansi --with dev --without ai,api,worker
+        poetry install --no-root --no-ansi --with dev --without api
```
gecBurton, line 54:
```suggestion
        poetry install --no-root --no-ansi
```

PR #710, opened 2024-07-02
Show citations in insertion order, still grouped by file.

brunns:
## Things to check

### DIFF
--- a/django_app/tests/test_views.py
+++ b/django_app/tests/test_views.py
@@ -372,6 +372,38 @@ def test_view_session_with_documents(chat_message: ChatMessage, client: Client):
 
 
 @pytest.mark.django_db()
+def test_citations_shown_in_correct_order(
+    client: Client, alice: User, chat_history: ChatHistory, several_files: Sequence[File]
+):
+    # Given
+    client.force_login(alice)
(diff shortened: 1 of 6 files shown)


django_app/tests/test_views.py
```diff
@@ -371,6 +371,38 @@ def test_view_session_with_documents(chat_message: ChatMessage, client: Client):
...
+    Citation.objects.create(file=several_files[0], chat_message=chat_message, text="Citation 5")
+
+    # When
+    response = client.get(f"/citations/{chat_message.id}/")
+
+    # Then
+    assert response.status_code == HTTPStatus.OK
+    soup = BeautifulSoup(response.content)
+    sources_panel = soup.select("div.iai-panel")[1]
+    files = sources_panel.find_all("h3")
```
rachaelcodes, line 395: Further to our discussion in Slack, would we be able to use something like semantic class names to make this clearer (and possibly less brittle)?

brunns: If @KevinEtchells is happy, so am I!

rachaelcodes (approved):
Looks like this does the trick. If we come to an agreement on how to select elements for tests, it would be good to get that in too for the tests you've added to `test_views.py`

PR #711, opened 2024-07-02
Fix scripts load order

KevinEtchells:
## Context

The TrustedTypes JavaScript was be loaded after other scripts that depend on it. This is a regression from an update yesterday.

* Load the TrustedTypes JavaScript earlier in the page, and tidy-up

## Guidance to review

The easiest way to check this is now working is existing chat messages will now be properly formatted on page load.


brunns (approved):
LGTM.

PR #713, opened 2024-07-02
Track chat routes on Plausible

rachaelcodes:
## Context
This allows us to track the route used in a chat message on Plausible in streamed chat.

Sends the route to the JS app for non-staff users in a new 'hidden-route' event type (this can be removed when the route is available to all users).

For both the 'route' and 'hidden-route' event types, trigger a custom `Chat-message-route-[ROUTE NAME]` event in Plausible: 
https://plausible.io/docs/custom-event-goals#trigger-custom-events-manually-with-a-javascript-function

## Guidance to review
- For someone with experience with Plausible (@KevinEtchells) : Does this look like it will work?
- Does the code run locally (without Plausible) without triggering console errors? 

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-439
https://plausible.io/docs/custom-event-goals#trigger-custom-events-manually-with-a-javascript-function

## Things to check

### DIFF
--- a/django_app/frontend/style.css
+++ b/django_app/frontend/style.css
@@ -8433,7 +8433,6 @@ body:has(.iai-environment-warning) main {
 .iai-chat-bubble__route {
   font-size: 0.625rem;
   line-height: 0.825rem;
-  display: none;
 }
 
 .iai-chat-bubble__sources-heading, .iai-chat-bubbles__sources-link {
--- a/django_app/frontend/js/chats/streaming.js
+++ b/django_app/frontend/js/chats/streaming.js
@@ -154,6 +154,12 @@ class ChatMessage extends HTMLElement {
                     route.textContent = message.data;
                     route.removeAttribute("hidden");
                 }
+
+                // send feedback to Plausible
+                let plausible = /** @type {any} */ (window).plausible;
+                if (typeof(plausible) !== 'undefined') {
+                    plausible(`Chat-message-route-${message.data}`);
+                }
             } else if (message.type === 'end') {
                 sourcesContainer.showCitations(message.data.message_id);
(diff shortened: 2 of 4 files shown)


brunns (approved):
Looks good for now.

rachaelcodes:
> If you feel this will be burdensome in future (e.g. if we're adding lots of routes, having to remember to add the custom events each time etc.) then it may be worth considering just having a single `route` event with additional props: https://plausible.io/docs/custom-props/for-custom-events#2-using-the-manual-method
> 
> Do we need the `hidden-route` message type in the `webSocket.onmessage` callback? I appreciate you've commented that this is only temporary, but as this is already hidden perhaps we don't need it at all? I'd be keen to not add any more complexity to this function than we need to, but my question may be due to my lack of understanding of the problem.

@KevinEtchells given that we sometimes add new routes for testing, I'm tempted to use the single 'route' approach you mentioned above

The difference between the `hidden-route` and the `route` event is that the `hidden-route` doesn't remove the 'hidden' attribute from the route object. The other option would be to pass the 'show_route' boolean from Python to JS, which I felt would be even more fiddly to unpick when we're removing it. Let me know if you think otherwise though.

django_app/frontend/style.css
```diff
@@ -8433,7 +8433,6 @@ body:has(.iai-environment-warning) main {
 .iai-chat-bubble__route {
   font-size: 0.625rem;
   line-height: 0.825rem;
-  display: none;
```
rachaelcodes, line 8436: This was removed as part of https://github.com/i-dot-ai/redbox-copilot/pull/698 and I think might have got re-added during a rebase

KevinEtchells (approved):

PR #716, opened 2024-07-03
Feature/build and deploy action

252afh:
## Context

The deploy-dev.yml action currently only deploys dev. Increase the surface area of this action to deploy to preprod and prod too.

- Added release trigger
- Added merge to main trigger
- Added a lot of logic to env var step to set env vars for all exection flows
- Changed checkout step to use calculated hash
- Added slack notification for success and failure

## Guidance to review

This workflow will trigger from a dispatch call for testing

## Things to check


.github/workflows/build-and-release.yml
```diff
@@ -1,6 +1,12 @@
 name: Deploy to dev
```
RyanWhite25, line 1: This'll need updating now I imagine 🙂

.github/workflows/build-and-release.yml
```diff
@@ -28,13 +34,59 @@ jobs:
...
       infra-config-repo: ${{ steps.export.outputs.infra-config-repo }}
-
+      environment: ${{ steps.env-var.outputs.environment }}
+      commit-hash: ${{ steps.env-var.outputs.commit-hash }}
+      author-name: ${{ steps.env-var.outputs.author-name }}
+      core-repo-reference: ${{ steps.env-var.outputs.core-repo-reference }}
+      config-repo-reference: ${{ steps.env-var.outputs.config-repo-reference }}
 
     steps:
+      - id: env-var
```
RyanWhite25, line 44: It looks to me like we might potentially be able to reduce this env-vars declaration through the use of the GitHub actions language. There's a lot going on here and on initial viewing I found it quite hard to dissect what the expected behaviour is. I'm wondering if it's worth us using the more-native GH actions syntax to do this if we can? It's hard to explain what I mean over GH comments, but do you mind if I have a go at a quick refactor of this on a branch of your branch to show you an example?

.github/workflows/build-and-release.yml
```diff
@@ -28,13 +34,59 @@ jobs:
...
+        run: |
+          # prod env from release
+          if [[ "${{ github.event_name }}" == "release" ]]; then
+            export ENVIRONMENT=prod
+            echo "::set-output name=commit-hash::$GITHUB_SHA"
```
RyanWhite25, line 49: Does the commit-hash var have to be echo'd out this way or can it be exported (as the other variables are) to keep it in-line with the other declarations?

.github/workflows/build-and-release.yml
```diff
@@ -86,17 +138,71 @@ jobs:
...
+      CORE_INFRA_REF: ${{ needs.set-vars.outputs.core-repo-reference }}
+      INFRA_CONFIG_REF: ${{ needs.set-vars.outputs.config-repo-reference }}
+      IMAGE_TAG: ${{ needs.set-vars.outputs.commit-hash }}
     secrets:
       AWS_GITHUBRUNNER_PAT: ${{ secrets.AWS_GITHUBRUNNER_PAT}}
       AWS_GITHUBRUNNER_PAT_USER: ${{ secrets.AWS_GITHUBRUNNER_PAT_USER }}
       AWS_REGION: ${{ secrets.AWS_REGION }}
       AWS_ACCOUNT_ID: ${{ secrets.AWS_ACCOUNT_ID }}
 
+  notify-slack:
```
RyanWhite25, line 152: Awesome! I love a good slack notification. We should look at turning this into a reusable GH action for sure, would be useful on pretty much every project I imagine. I can raise a ticket if you haven't already and we can bake it into the cookiecutter.

RyanWhite25 (commented):
Nice one - there's a lot of logic to bake in here and you've managed to squeeze it all in, I can see how this can be a pain to get right! Looking good with a few non-blocking comments. Let me know your thoughts and I can approve afterwards 👍

.github/workflows/_cicd.yml
```diff
@@ -0,0 +1,57 @@
...
+    with:
+      environment: dev
+      core_infra_repo_reference: ${{ inputs.core_repo_reference }}
+      config_repo_reference: ${{ inputs.config_repo_reference }}
+      commit_hash: ${{ inputs.commit_hash }}
+    secrets: inherit
+
+  release-preprod:
+    if: github.event_name == 'push' && github.ref_name == 'main'
+    uses: i-dot-ai/redbox-copilot/.github/workflows/build-and-release.yml@a37b08ab5253a8ba3b592b61cb367c47d022f298
```
RyanWhite25, line 41: Should we unpin this now?

.github/workflows/_cicd.yml
```diff
@@ -0,0 +1,57 @@
...
+    with:
+      environment: preprod
+      core_infra_repo_reference: 'refs/heads/main'
+      config_repo_reference: 'refs/heads/main'
+      commit_hash: ${{ github.sha }}
+    secrets: inherit
+
+  release-prod:
+    if: github.event_name == 'release'
+    uses: i-dot-ai/redbox-copilot/.github/workflows/build-and-release.yml@a37b08ab5253a8ba3b592b61cb367c47d022f298
```
RyanWhite25, line 51: Should we unpin this now?

.github/workflows/_cicd.yml
```diff
@@ -0,0 +1,57 @@
...
+      config_repo_reference: ${{ inputs.config_repo_reference }}
+      commit_hash: ${{ inputs.commit_hash }}
+    secrets: inherit
+
+  release-preprod:
+    if: github.event_name == 'push' && github.ref_name == 'main'
+    uses: i-dot-ai/redbox-copilot/.github/workflows/build-and-release.yml@a37b08ab5253a8ba3b592b61cb367c47d022f298
+    with:
+      environment: preprod
+      core_infra_repo_reference: 'refs/heads/main'
```
RyanWhite25, line 44: Nice - I prefer this hard-coding over the `env:` block as I find it expresses intent a lot better 👍

RyanWhite25 (approved):
Looks good

.github/workflows/_cicd.yml
252afh: I found that you actually can't use `env` defined vars in a `with` block, caused me headaches for ages lol

PR #718, opened 2024-07-03
Add posthog

KevinEtchells:
## Context

Adding Posthog analytics to collect data we are unable to using Plausible. Currently only for pre-prod until we've tested there and are happy to move it to prod.

Due to our Content Security Policy, we've needed to bundle all of the Posthog scripts at build time, rather than just adding the script in the usual way.

## Guidance to review

* Check the code looks sensible
* Testing will commence on pre-prod once merged and deployed
* You could temporarily remove the environment check to see that it loads locally
* @Samuel-Hoskin has confirmed data is coming into the dashboard when we tested


gecBurton (approved):
looks sensible

PR #739, opened 2024-07-08
Feature/redbox 430 backwards compatiblity

jamesrichards4:
## Context

We don't want to assume that all deployments of redbox will use Azure for embeddings.

Make the embeddings objects conditionally using env vars. We can switch between OpenAI and Azure. The OpenAI embeddings class should allow us to use a deployment of LocalAI (example in docker-compose) to use sentencetransformers to create embeddings, which is as we were before the move to Azure. This means all embeddings are done remotely, but other deployments don't need to do a breaking change, they can deploy LocalAI alongside instead

## Guidance to review

I can't test the docker compose setup works properly due to the LocalAI images not being distributed for ARM arch. I did test using the default embedding model in the aio images but this won't deliver the backwards compatibility we want as this model is fixed. If possible we should test that the message commented in the docker-compose file works and allows using the old sentencetransformer model

I've added a whole bunch of vscode stuff. Be good to check this out with others to make sure it works nicely.

I've added integration env as a base for docker compose containers. Overrides in .env apply but we now get test env for unit tests and integration env for started containers as a baseline. Shout if this is wrong or causes issues. 
NB we should probably move .env to be service specific so core-api uses .env.integration and core-api/.env optionally

## Things to check


Makefile
```diff
@@ -106,6 +106,12 @@ reset-db:  ## Reset Django database
 	docker compose down db --volumes
 	docker compose up -d db
 
+.PHONY: reset-elastic
+reset-elastic:  ## Reset Django database
```
gecBurton, line 110: 👍

core-api/core_api/dependencies.py
```diff
@@ -149,15 +142,12 @@ def get_llm(env: Annotated[Settings, Depends(get_env)]) -> ChatLiteLLM:
 
         llm = ChatLiteLLM(
             model=env.azure_openai_model,
+            azure_api_key=env.azure_openai_api_key,
             streaming=True,
             api_base=env.azure_openai_endpoint,
             max_tokens=env.llm_max_tokens,
             callbacks=[logger_callback],
         )
-    elif env.anthropic_api_key is not None:
```
gecBurton, line 157: 👍 good to get rid of this

core-api/core_api/semantic_routes.py
```diff
@@ -149,9 +149,20 @@ def get_semantic_routes():
 
 
 def get_semantic_routing_encoder(env: Annotated[Settings, Depends(get_env)]):
-    return AzureOpenAIEncoder(
-        azure_endpoint=env.azure_openai_endpoint, api_version="2023-05-15", model=env.azure_embedding_model
-    )
+    """
+    TODO: This is a duplication of the logic for getting the LangChain embedding model used elsewhere
+    We should replace semanticrouter with our own implementation to avoid this
```
gecBurton, line 154: good point

core-api/pyproject.toml
```diff
@@ -35,6 +35,8 @@ litellm = "^1.41.11"
 pytest = "^8.2.2"
 pytest-cov = "^5.0.0"
 pytest-asyncio = "^0.23.6"
+pytest-dotenv = "^0.5.2"
+mypy = "^1.9.0"
```
gecBurton, line 39: not a blocker but i dont think mypy belongs here

core-api/pyproject.toml
```diff
@@ -44,3 +46,15 @@ deepeval = "^0.21.64"
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.pytest.ini_options]
```
gecBurton, line 50: 👍

core-api/pyproject.toml
jamesrichards4: good point, missed this in a merge from main

gecBurton (approved):
this works for me locally

non-blocking, pls can we add a short into into the developer setup 

```
EMBEDDING_OPENAI_BASE_URL=http://embeddings:8080/v1
EMBEDDING_BACKEND=openai
EMBEDDING_DOCUMENT_FIELD_NAME=something
```

PR #740, opened 2024-07-08
CHORE: Added ned values to the example.env which I needed before I co…

JLoweAI:
## Context

When I was trying to get the API working locally, the .env.example didn't contain the relevant (blank) values linked to Azure which I needed to get it working 

Changes to .example.env


gecBurton (approved):
LGTM , thank you

PR #743, opened 2024-07-09
Implement new route-display design

KevinEtchells:
## Context

Update the route display to match the latest designs. This is the first step to making this visible to all users.

* Update the design
* Create a new tooltip component
* Update to the lastest i.AI Design System version

screenshot showing the new tooltip design

## Guidance to review

* Check the route displays okay
* That the tooltip appears on mouse-over, and disappears on mouse-leave is removed
* This also works on focus (for accessibility) and click (for touch-screens)


brunns (approved):
Belay that - had to hard-refresh my browser.

PR #746, opened 2024-07-09
chore/faster-mypy

gecBurton:
## Context

As an Engineer I want mypy to be run once over all the code, this will make it much faster in CI

1. all instances of mypy have been removed except at the top level
2. some unused dependencies have been removed

## Guidance to review

Note that in this PR the mypy check takes 17 seconds whereas the last run in main took more than 3 minutes

## Things to check


brunns (approved):
I'd probably have gone with pyright, but looks good.

PR #751, opened 2024-07-09
Changed slack integration to use webhook url instead of bot token

252afh:
## Context

We couldn't get the slack bot token back, so I've changed the slack integration to use the webhook url instead from a slack app to make it recoverable.

- Remove channel-id
- Change slack bot token to webhook url

## Things to check


gecBurton (approved):
LGTM

PR #768, opened 2024-07-11
chore/increase storage to 30 gb for core and worker

gecBurton:
## Context

As an Engineer I want to increase the storage for the core-api and worker 

## Things to check


brunns (approved):
LGTM.

PR #770, opened 2024-07-11
REDBOX-440 - Editable chat title

brunns:
## Things to check


tests/test_journey.py
```diff
@@ -113,6 +113,13 @@ def test_user_journey(page: Page, email_address: str):
         latest_chat_response = chats_page.wait_for_latest_message()
         assert latest_chat_response.text
         assert latest_chat_response.route.startswith(route)
+
+    # Navigate to old chat & rename
+    chats_page = chats_page.navigate_to_titled_chat("What architecture is in use?")
+    chats_page.chat_title = "About tech stuff."
+    chats_page = chats_page.start_new_chat()
+    chats_page = chats_page.navigate_to_titled_chat("About tech stuff.")
```
rachaelcodes, line 121: I really like the tests - it's v clear what you've changed

django_app/frontend/src/js/chats/streaming.js
```diff
@@ -297,3 +307,125 @@ class DocumentSelector extends HTMLElement {
   }
 }
 customElements.define("document-selector", DocumentSelector);
+
```
rachaelcodes, line 310: Question: do we want to start thinking about breaking this file up into its different components?

brunns: That's not a bad idea at all. The feedback component is already in a file of its own.

django_app/frontend/src/js/chats/streaming.js
```diff
@@ -297,3 +307,125 @@ class DocumentSelector extends HTMLElement {
...
+    if (this.input && this.heading) {
+      this.input.value = e.detail.title;
+      this.heading.textContent = `${e.detail.title}`;
+      this.heading.classList.remove("govuk-visually-hidden");
+    }
+  };
+
+  update = () => {
+    const newTitle = this.input?.value;
+    console.log(`updating chat title to "${newTitle}"`);
```
rachaelcodes, line 409: are we keeping this for the user?

brunns: Good catch - I'll get rid.

brunns: Gone.

django_app/frontend/src/js/chats/streaming.js
KevinEtchells: I think that's a good idea, and now we have Parcel doing the bundling this should be relatively clean and simple.

PR #772, opened 2024-07-12
Adds a pyproject to the notebooks directory

ghost:
## Context

The new service structure makes notebooks hard. This makes them easy again. Notebooks now have a separate poetry environment that loads all the others.

* Adds a notebook `pyproject.toml` with all other services as dependencies
* Updates `quickupload.ipynb` to use them to show they work
* Updates `quickupload.ipynb` to use the new embedding model
* Adds a script to `quickupload.ipynb` to quickly embed and save a directory of raw files to `.jsonl` without using the vector store

## Guidance to review

* Confirm you're happy with the new pyproject

## Things to check

- [x] [I have run integration tests](https://github.com/i-dot-ai/redbox/actions/runs/9904412945)


jamesrichards4 (approved):
Just noting that at some point this might fail as we may end up with incompatible dependencies between worker and core-api but this seems simplest until then

PR #780, opened 2024-07-15
Improve focus behaviour of chat messages and feedback

KevinEtchells:
## Context

A couple of improvements primarily for keyboard-only and screen-reader users

* The feedback functionality is split into 4 sections. When a new section is shown, move focus to this new section. This means focus isn't "lost", and users can perceive where they are.

* When a new message is being streamed, focus was moving to the container for the new message and feedback buttons. This is now just moving to the new message - which looks tidier for all users.

## Guidance to review

* To test, you will need to just use keyboard, not mouse. `Tab` to go forwards. `Shift` + `Tab` to go backwards. `Space` to interact with control.
* You should see the focus indicator outlines as you navigate through.

### DIFF
--- a/django_app/frontend/src/js/chats/feedback.js
+++ b/django_app/frontend/src/js/chats/feedback.js
@@ -15,7 +15,7 @@ class FeedbackButtons extends HTMLElement {
     `;
 
     this.innerHTML = `
-        <div class="feedback__container feedback__container--1">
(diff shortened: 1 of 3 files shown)


rachaelcodes (approved):
Nice and easy to use with a keyboard ⌨️

PR #792, opened 2024-07-16
Announce when message streaming has completed to screen-reader users

KevinEtchells:
## Context

When testing with a screen-reader, I noticed it was announced when a new message is loading, but not when it has completed - leaving users to guess when it's complete.

* Announce "Response complete" to screen-reader users when a response has finished streaming

## Guidance to review

* Check the code looks sensible
* If you can't test with a screen-reader, I understand! And I'd be happy to give you a demo, if you want. But I'm happy it's working.


brunns (approved):
LGTM.

PR #796, opened 2024-07-16
Refactor/rename django folder

lcardno10:
## Things to check


gecBurton (approved):
thanks, this has been one of those things that Ive always wanted to do but never quite found the energy!

PR #804, opened 2024-07-18
Replace nul fields in user entered text.

brunns:
## Things to check


gecBurton (changes requested):
one quibble, we have three instances of:

```python
self.text = self.text.replace("\x00", "\ufffd") if self.text else self.text
```

please could we have a function to hanlde this with a brief docstring as i think that this bug is sufficiently niche as to need some explanation.

otherwise LGTM

gecBurton (approved):

PR #806, opened 2024-07-18
Fix journey tests for updated routes.

brunns:
## Things to check


andy-symonds (approved):
I like the refactor to check `startswith(route)`. Looks good to me.

PR #810, opened 2024-07-18
Feature/log failure to ingest code

gecBurton:
## Context

As a User I want to know that the ingest process has failed to extract any text

Using the document in the ticket I see that `unstructured.partition` returns 0 elements without error or warning.

From the perspective of the user this is an error, so I am now raising an `Exception` so that this is caught by the worker and the file is marked as `failed`

The test has been extended to cover a manually corrupted pdf

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-468

## Things to check


redbox-core/redbox/loader/file_loader.py
```diff
@@ -40,6 +40,9 @@ def lazy_load(self) -> Iterator[Document]:  # <-- Does not take any arguments
         to yield documents one by one.
         """
         elements = partition(file=self.file_bytes, strategy=self.env.partition_strategy)
+        if not elements:
```
gecBurton, line 43: this is the key change

PR #815, opened 2024-07-19
chore/simplification

gecBurton:
## Context

less is more

## Things to check


worker/tests/test_app.py
```diff
@@ -55,35 +54,13 @@ async def test_ingest_file(
...
-                                            "metadata.creator_user_uuid.keyword": str(file.creator_user_uuid),
-                                        }
-                                    },
-                                ]
-                            }
-                        }
-                    },
-                )
-            )
+            chunks = storage_handler.get_file_chunks(file.uuid, file.creator_user_uuid)
```
jamesrichards4, line 65: I'm hoping to remove all chunks functions from the storage_handler then remove those functions, the API functions and the Chunk class

worker/tests/conftest.py
```diff
@@ -30,22 +28,3 @@ def s3_client(env: Settings):
...
     return env.elasticsearch_client()
-
-
-@pytest.fixture(scope="session")
-def es_index(env: Settings) -> str:
-    return f"{env.elastic_root_index}-chunk"
-
-
-@pytest.fixture(autouse=True, scope="session")
-def create_index(env: Settings, es_index):
```
jamesrichards4, line 41: This means we're using a clean test index each run which is nice?

worker/tests/test_app.py
gecBurton: accepted

PR #824, opened 2024-07-22
feature/pin-redbox-core-dependency-for-worker

gecBurton:
## Context

As an Engineer I want to pin the worker and core-api's dependency on redbox-core to a specific version so that I do not have to rebuild to worker/core-api every time an unrelated change gets made in core

1. every thing that uses `Unstructured` and therefore `torch` has been moved to `worker`
2. redbox-core has been pinned to a specific version in the `worker`
3. I also had to fix some typing

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-548

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox/actions/runs/10041040900


jamesrichards4 (commented):
When we're developing wouldn't this change make it so we don't see changes in redbox-core in core-api/worker? Feels like we want develop install for dev and only a fixed install for docker builds somehow?

gecBurton:
> When we're developing wouldn't this change make it so we don't see changes in redbox-core in core-api/worker? Feels like we want develop install for dev and only a fixed install for docker builds somehow?

good point.... ill have a think about this

PR #829, opened 2024-07-23
Remove makefile targets and add ecs scaledown

252afh:
## Context

ECS services running on the weekend and at night shouldn't scale even under demand as it's unlikely to be legitimate traffic. Change the ECS tasks to include a scheduled action that scales down and up depending on peak times

- Added bool to tell each ECS service to scale

## Guidance to review

Check the resources exist in terraform (currently no way to view in aws console)

## Things to check


brunns (approved):
LGTM.

PR #833, opened 2024-07-23
REDBOX-551 - Collect name and ai experience level

brunns:
## Context

https://technologyprogramme.atlassian.net/browse/REDBOX-551

## Things to check


django_app/frontend/src/styles.scss
```diff
@@ -433,3 +433,39 @@ button.iai-chat-bubble__citations-button {
...
+.govuk-radios__item:has(.govuk-radios__hint) .govuk-radios__label::after {
+  top: 14px;
+}
+.govuk-radios__input {
+  height: 1.5rem;
+  width: 1.5rem;
+}
+.govuk-radios__label,
+.govuk-radios__hint {
+  font-size: 0.75rem;
```
rachaelcodes, line 465: This is very small onscreen

django_app/redbox_app/redbox_core/models.py
```diff
@@ -94,6 +94,25 @@ class Profession(models.TextChoices):
         VET = "VET", _("Veterinary")
         OT = "OT", _("Other")
 
+    class AIExperienceLevel(models.TextChoices):
```
rachaelcodes, line 97: How sure are we that we're going to keep these categories and labels? It'll be annoying to update them in the future.

rachaelcodes (approved):
Potentially gives us a small amount tech debt with the hardcoded AIExperienceLevel labels, and we might want to get some user feedback on the font size.

However, given the design/content has been approved and the release deadline, I don't think these should block merging at this point.

django_app/redbox_app/redbox_core/models.py
brunns: I did talk to Charlotte about this. She said that the categories would be fairly stable, so I decided on an enumeration over a foreign key. If it turns out they do change often, we might reverse that decision in future.

PR #835, opened 2024-07-23
[REDBOX-490] Calculate summarisation chunk size from context size

andy-symonds:
## Context

We want to calculate stuff chunk size as a ratio of context window size, as they will always be linked and this prevent errors in a mismatch of values when. being deployed

Renamed `summarisation_chunk_max_tokens` to `stuff_chunk_max_tokens` as this env var has more to do with stuffing than a specific summarisation use case

## Guidance to review

1. I have updated `infrastructure/data.tf` and `infrastructure/variables.tf` --> should we also remove the default value of `stuff_chunk_max_tokens` in `infrastructure/variables.tf` @gecBurton @jamesrichards4 ?

2. `.env.test` has been updated so the instead of using `AI__STUFF_CHUNK_MAX_TOKENS` and setting it to a small value, as this is now calculated as a ratio of context_window_size, `.env.test` now has an env var `AI__STUFF_CONTEXT_RATIO == 0.01` (i.e. small), in order to trigger map_reduce.

## Things to check


redbox-core/redbox/models/settings.py
```diff
@@ -118,14 +118,18 @@ class AISettings(BaseModel):
...
     map_system_prompt: str = MAP_SYSTEM_PROMPT
     map_question_prompt: str = MAP_QUESTION_PROMPT
     map_document_prompt: str = MAP_DOCUMENT_PROMPT
     reduce_system_prompt: str = REDUCE_SYSTEM_PROMPT
     reduce_question_prompt: str = REDUCE_QUESTION_PROMPT
 
+    @computed_field  # type: ignore
+    @property
+    def stuff_chunk_max_tokens(self) -> int:
+        return int(self.context_window_size * 0.8)
```
jamesrichards4, line 131: could we make the 0.8 another setting? stuff_chunk_context_ratio which defaults to 0.75

andy-symonds:
@gecBurton @jamesrichards4 - In `variables.tf` can I safely remove the definition of `stuff_chunk_max_tokens` as it is now a computed field or for AWS does it need to stay there?

I will at least remove the default value

jamesrichards4 (approved):

PR #838, opened 2024-07-24
File reingestion management command

rachaelcodes:
## Context
Now that we're running management tasks through ECS commands, this gives us a new option for running reingestion to core-api. Using Django admin for this task caused timeout problems if done in batches that were too large. If successful, I will remove the old Admin command in a later PR

* Adds and tests a new 'reingest files' management command.
* updates the expired file deletion management command using the new common `INACTIVE_STATUSES` constant
* updates the expired file deletion tests so that the requests mock takes into account other matching files - avoiding annoying errors when testing locally.

## Guidance to review
Does the command and its test make sense?

## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/management/commands/reingest_files.py
+++ b/django_app/redbox_app/redbox_core/management/commands/reingest_files.py
@@ -19,10 +19,7 @@ class Command(BaseCommand):
 
     def handle(self, *_args, **_kwargs):
         self.stdout.write(self.style.NOTICE("Reingesting active files from Django"))
-        counters = {
-            "success": 0,
-            "errors": 0,
-        }
+        successes, errors = 0, 0
 
         for file in File.objects.exclude(status__in=INACTIVE_STATUSES):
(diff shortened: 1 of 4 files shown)


django_app/redbox_app/redbox_core/management/commands/reingest_files.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Command(BaseCommand):
+    help = """This is an ad-hoc command when changes to the AI pipeline (e.g. a new embedding strategy)
+    mean we need to regenerate chunks for all the current files.
+
+    It attempts to reupload file data to core-api for reingestion."""
+
+    def handle(self, *_args, **_kwargs):
+        self.stdout.write(self.style.NOTICE("Reingesting active files from Django"))
+        counters = {
```
gecBurton, line 22: v minor point but im not sure what this dict is adding? could just be:

```python
success, errors = 0, 0
```

?

gecBurton (approved):
one piece of nitpicking otherwise LGTM

PR #840, opened 2024-07-24
Make select_document message more general

andy-symonds:
## Context

If NoDocumentsSelected exception is triggered, currently the message the user receives is specific to summarisation, however, this can happen in other situations, so we need to make it more general.

Minor change to `SELECT_DOCUMENT` message.

## Things to check


brunns (approved):
LGTM.

PR #841, opened 2024-07-24
Fix errors on file upload.

brunns:
## Context

https://technologyprogramme.atlassian.net/browse/REDBOX-566

The ['File' model in 'redbox-core/redbox/models/file.py'](https://github.com/i-dot-ai/redbox/blob/main/redbox-core/redbox/models/file.py#L28) is optional - and for a short period we indeed get `None` back from Elastic. We need to accommodate this in `core-api` and `django-app`.

## Things to check


gecBurton (approved):
thx for the diagnosis - this has been baffling me for a while!

PR #843, opened 2024-07-24
feature/removed stuff_chunk_max_tokens

gecBurton:
## Context

As an Engineer I want to remove unused variables 

## Things to check


.env.test
```diff
@@ -3,8 +3,6 @@
 LLM_MAX_TOKENS=1024
 
 # === AI ===
-# AI__STUFF_CHUNK_MAX_TOKENS=80
-AI__STUFF_CHUNK_CONTEXT_RATIO=0.01
```
andy-symonds, line 7: @gecBurton we need to keep `AI__STUFF_CHUNK_CONTEXT_RATIO` - this is used along with context window size to calculate `stuff_chunk_max_tokens`. Withouth this ratio variable in .env.text the value of stuff_chunk_max_tokens` in the tests to be small enough to test the switch between large and small document summarisation.

gecBurton: my bad!

andy-symonds: Okay, I see the summarisation tests with large and small chunks have been removed (don't we still want them @jamesrichards4 as summarisaion route it still an option?)

andy-symonds (approved):

PR #844, opened 2024-07-24
Copy AI responses both formatted and unformatted.

brunns:

Make the copy button copy both formatted and unformatted text, so you can paste into Word or similar and get formatted text, or into a plain text editor and get plain text without Markup.

## Things to check


KevinEtchells (approved):
Nice one. That works better.

PR #849, opened 2024-07-24
Split django ecs task and added up/down scheduling

252afh:
## Context

The management commands we're doing could do with going on a schedule

I'm not sure how effective this will be at stopping them, `desired_count` may override it, but it's hard to test because of not being able to see the schedules in aws console

- Changed `django_command` variable to be an object to create each task from
- Updated consuming resources of django-command module
- Added var to pass through to core infra

## Things to check


252afh:
No longer needed, ECS django commands have been moved to lambdas

PR #869, opened 2024-07-26
Fix status updates on file upload and reingestion

rachaelcodes:
## Context
When we ran the re-ingestion task, file status appeared as 'unknown' in the front end. While digging into this, I also noticed one of the core-api file statuses was not included in the Django app.

* Adds 'processing' from redbox-core to the Django app
* Sets the file status to 'processing' both on file upload and reingestion. 

## Guidance to review
Are the changes and test reasonable?

## Things to check


brunns (approved):
All looks good.

PR #870, opened 2024-07-26
Allow user to override auto-scrolling

KevinEtchells:
## Context

When a new message streams, the latest chunk is automatically scrolled into view (to prevent it from being hidden behind the sticky message-input). Feedback has shown users want to control their own scrolling behaviour rather than having to wait for streaming to finish.

* Detect if a user has scrolled during streaming
* If so, stop the auto-scrolling

## Guidance to review

Test a few messages. Both with and without scrolling. Is the behaviour what you would expect?

## Things to check


brunns (approved):
Works for me.

PR #875, opened 2024-07-26
Spike/langgraph

jamesrichards4:
## Context

We want to move to a more capable AI framework and concentrate all our AI work in the redbox library redbox-core

Use LangGraph to run the computation rather than manually executing chains.
Move all chains/AI code to redbox-core and make it available as a single graph which can be imported
Add new tests which allow easy creation of scenarios

## Guidance to review

The current solution should be unchanged. This brings the new LangGraph code in alongside for now.

## Things to check


redbox-core/redbox/graph/chat.py
```diff
@@ -0,0 +1,157 @@
...
+    app = StateGraph(ChatMapReduceState)
+
+    app.add_node(
+        "llm_map", build_llm_map_chain(llm, tokeniser, env, env.ai.map_system_prompt, env.ai.chat_map_question_prompt)
+    )
+    app.add_node(
+        "reduce",
+        build_reduce_docs_step(
+            TokenTextSplitter(
+                model_name="gpt-4",
```
lmwilkigov, line 148: Should this be `gpt-4o` instead?

notebooks/langgraph.ipynb
lmwilkigov, line 1: Love the graph visualisation! Could we also serialise the images to the `docs/assets` directory so we can utilise them in the docs?

lmwilkigov (approved):
Very clean code 🙌. My main thought is that I'm likely not fluent in LangGraph enough to spot anything needing a change. I'd be keen to get this up in Dev to test it out and then start documenting the Graph quite heavily

redbox-core/redbox/graph/chat.py
jamesrichards4: yeah, I'm not sure, this should be an env var so we switch it as needed

PR #887, opened 2024-07-30
moved test data to tests/

gecBurton:
## Context

As an Engineer I want test artefacts to be defines within the tests directory

## Things to check


jamesrichards4:
the data classes used here are also used in core-api so this doesn't work. They provide a standard way to create the test data which I think is useful

PR #894, opened 2024-08-01
AISettings no longer part of Settings

gecBurton:
## Context

This PR is part of https://technologyprogramme.atlassian.net/browse/REDBOX-572

it is a PR in its own right. but is not the fully story

in this PR i co-opt `AISettings` to be the dynamic, per-request settings and I leave the existing `Settings` as the static system-level-settings. overtime we will:

* configure the django app to set the `AISettings` in the request
* move more settings out of the `Settings` and into the `AISettings` 
* tidy up

1. `AISettings` is no longer part of `Settings`, i.e. it is not set by the env vars
2. `AISettings` is now part of `ChatMessage`, i.e. it may be set by the django app (but accepts None at the moment)

## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox/actions/runs/10212335819


tests/pages.py
```diff
@@ -253,7 +253,9 @@ def all_documents(self) -> list[DocumentRow]:
     def _doc_from_element(element: Locator) -> DocumentRow:
         filename = element.locator(".iai-doc-list__cell--file-name").inner_text()
         status = element.locator(".iai-doc-list__cell--status").inner_text()
-        completed = element.evaluate("element => element.closest('.iai-doc-list').classList.contains('iai-doc-list--complete')")
```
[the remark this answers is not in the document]
gecBurton: agreed, `make format` did it form me.. 🤷‍♂️

jamesrichards4:
We're returning to making our components every request again here which feels like a regression. I was thinking we could use https://langchain-ai.github.io/langgraph/how-tos/configuration/#configure-the-graph and use the AISettings as the config schema but that doesn't work with the ElasticsearchRetrievers from a quick look.

Does this AISettings object belong in the query/request object (ChainInput)? That way core-api can just pass the whole request to redbox and we can update the retrievers etc to read values from state["query"].settings as needed? @wpfl-dbt  thoughts on expanding the request object to cover settings?

gecBurton:
> We're returning to making our components every request again here which feels like a regression

is this not the point?

core-api/core_api/dependencies.py
```diff
@@ -38,23 +37,11 @@ def get_parameterised_retriever(
 
     Runnable returns a list of Chunks.
     """
-    default_params = {
```
gecBurton, line 41: all moved to AISettings

redbox-core/redbox/app.py
```diff
@@ -55,14 +55,22 @@ def __init__(
         app.add_node(
             ChatRoute.search,
             get_search_graph(
-                _llm, _parameterised_retriever.with_config(tags=[Redbox.SOURCE_DOCUMENTS_TAG]), _tokeniser, _env, debug
+                _llm,
+                _parameterised_retriever.with_config(tags=[Redbox.SOURCE_DOCUMENTS_TAG]),
+                _tokeniser,
+                env.llm_max_tokens,
```
gecBurton, line 61: there is a strong case for moving llm_max_tokens to the `AISettings` however this is also used in `get_llm` which is set during the construction of Redbox

redbox-core/redbox/chains/components.py
```diff
@@ -67,21 +66,9 @@ def get_parameterised_retriever(env: Settings, embeddings: Embeddings | None = N
 
     Runnable returns a list of Chunks.
     """
-    default_params = {
```
gecBurton, line 70: all moved to AISettings

redbox-core/redbox/models/settings.py
```diff
@@ -156,8 +37,6 @@ class ElasticCloudSettings(BaseModel):
 class Settings(BaseSettings):
     """Settings for the redbox application."""
 
-    ai: AISettings = AISettings()
```
gecBurton, line 159: this is a key change

redbox-core/redbox/app.py
jamesrichards4: this makes sense to me. We should be able to get this out of the constructor level to make this work in a follow up PR. More generally I think these kind of vars want to be configurable in the graph as we think about things being more agentic.

jamesrichards4 (approved):
We could maybe chat about the max_tokens thing on a follow up if needed but this looks great

PR #898, opened 2024-08-01
Add comment explaining why poetry-plugin-bundle is pinned

brunns:
## Context

Add comment explaining that poetry-plugin-bundle is pinned until https://github.com/python-poetry/poetry-plugin-bundle/issues/112 is fixed.

## Things to check


rachaelcodes (approved):
another good comment

PR #900, opened 2024-08-01
Setting embedding batch size. Fixing some vscode env nonsense too

jamesrichards4:
## Context

We hit rate limiting on large files due to embedding too fast.

Set the embedding batch size on Azure Embeddings object (missed in previous work). This should drop the rate at which we're consuming our rate limit and smooth things out for large files where there are a large number of documents queued up

## Guidance to review

The vscode stuff is separate but bothering me locally and broke testing

## Things to check


gecBurton (approved):
good spot!

mypy errors should disappear on merge

PR #906, opened 2024-08-02
feature/moved ai models to ai_settigs

gecBurton:
## Things to check

### DIFF
--- a/redbox-core/redbox/chains/components.py
+++ b/redbox-core/redbox/chains/components.py
@@ -12,6 +12,7 @@ from redbox.api.callbacks import LoggerCallbackHandler
 from redbox.models.chain import AISettings
 from redbox.models.settings import Settings
 from redbox.retriever import AllElasticsearchRetriever, ParameterisedElasticsearchRetriever
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
 
 log = logging.getLogger()
 
@@ -47,6 +48,8 @@ def get_llm(ai_settings: AISettings) -> ChatLiteLLM:
             max_tokens=ai_settings.llm_max_tokens,
             callbacks=[logger_callback],
         )
+    elif ai_settings.chat_backend == "fake":
+        return GenericFakeChatModel(messages=iter(["test_case.test_data.expected_llm_response"]))
     else:
         msg = "Unknown LLM model specified or missing"
         log.exception(msg)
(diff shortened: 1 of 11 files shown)


redbox-core/redbox/models/chain.py
```diff
@@ -128,14 +128,22 @@ class AISettings(BaseModel):
...
-    llm_max_tokens: int = 1024
 
-    # size: int = 19 rag_k
     match_boost: int = 1
-    # num_candidates: int = 13 rag_num_candidates
     knn_boost: int = 1
     similarity_threshold: int = 0
 
+    chat_backend: Literal["azure", "openai", "fake"] = "azure"
+    fake_backend_responses: list = ["hello"]
```
gecBurton, line 137: I hate me too... but I cant think of another way. all ideas welcome

redbox-core/redbox/chains/components.py
```diff
@@ -1,20 +1,60 @@
...
-        api_key=convert_to_secret_str(env.azure_openai_api_key),
-        azure_endpoint=env.azure_openai_endpoint,
-        model=env.azure_openai_model,
-    )
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
+
+log = logging.getLogger()
+
+
+def get_llm(ai_settings: AISettings) -> ChatLiteLLM:
```
jamesrichards4, line 20: We shouldn't make the LLM object every time we need one. Adding functools.lru_cache(4) should means we only get a different one if the settings change up to some max (4 here).

Might have to set AISettings to be frozen to be hashable for lru_cache to work but that's what we want anyway

redbox-core/redbox/chains/graph.py
```diff
@@ -100,19 +100,25 @@ def set_prompt_args(state: ChainState):
...
-    )
+    @chain
+    def _build_llm_chain(chain_state: ChainState):
+        llm = get_llm(chain_state["query"].ai_settings)
+        _llm = llm.with_config(tags=["response_flag"]) if final_response_chain else llm
+
+        return RunnableParallel(
+            {
+                "response": make_chat_prompt_from_messages_runnable(
+                    tokeniser=tokeniser, llm_max_tokens=chain_state["query"].ai_settings.llm_max_tokens
```
jamesrichards4, line 114: if max tokens is in the state can't we just get it inside this function and avoid passing it separately?

redbox-core/redbox/models/chain.py
jamesrichards4: Could we have the fake llm expect an env var and read it's responses from there? In tests we could set that var for the test/session as needed

PR #909, opened 2024-08-02
Remove chunk model

jamesrichards4:
## Context

We no longer use chunks so we should tidy up the remaining references

Remove the get_chunks endpoint in the file app
Remove chunk actions from storage_handler. It no longer maps chunks or is aware of the parent_file_uuid linking. Storage Handler now only manages files (but can manage other types in future as needed)
Remove Chunk and Metadata models and tests

## Things to check


redbox-core/redbox/storage/storage_handler.py
```diff
@@ -12,7 +12,7 @@ class BaseStorageHandler(ABC):
     """
 
     # dict comprehension for lowercase class name to class
-    model_type_map: ClassVar = {v.__name__.lower(): v for v in [Chunk, File]}
+    model_type_map: ClassVar = {v.__name__.lower(): v for v in [File]}
```
gecBurton, line 15: this makes me wonder what the point of storing the file is at all.... what information does it hold that django doesnt already know about?

redbox-core/redbox/storage/elasticsearch.py
```diff
@@ -122,22 +93,20 @@ def delete_items(self, items: list[PersistableModel]) -> ObjectApiResponse | Non
             body={"query": {"terms": {"_id": [str(item.uuid) for item in items]}}},
         )
 
+    def delete_user_items(self, model_type: str, user_uuid: UUID, filters: list[dict] = None) -> ObjectApiResponse | None:
+        target_index = f"{self.root_index}-{model_type.lower()}"
+        return self.es_client.delete_by_query(
+            index=target_index,
+            body=ElasticsearchStorageHandler.get_query_match_all_for_user(user_uuid, filters),
```
gecBurton, line 100: maybe for another PR: is there a `langchain_elasticsearch.ElasticsearchStore` method that does this?

redbox-core/redbox/storage/elasticsearch.py
```diff
@@ -122,22 +93,20 @@ def delete_items(self, items: list[PersistableModel]) -> ObjectApiResponse | Non
...
+        )
+
     def read_all_items(self, model_type: str, user_uuid: UUID) -> list[PersistableModel]:
```
gecBurton, line 103: maybe for another PR: is there a `langchain_elasticsearch.ElasticsearchStore` method that does this?

redbox-core/redbox/models/file.py
```diff
@@ -43,107 +42,6 @@ def __hash__(self):
         return hash(self.text) ^ hash(self.url) ^ hash(self.start_index)
 
 
-class Metadata(BaseModel):
```
gecBurton, line 46: delighted to see this go!

gecBurton (approved):
much clearer than i feared, pls merge when ready!

redbox-core/redbox/storage/storage_handler.py
jamesrichards4: it records the ingest status and in future likely more detail on that (which tools are ready, which version of the ingest process a file went through), that feels like it makes sense in core-api rather than then user facing app

redbox-core/redbox/storage/elasticsearch.py
jamesrichards4: yes I think there is another PR like this, using retrievers or ElasticsearchStore to do more of this

PR #912, opened 2024-08-05
Feature/ecs ready check and lambda image change

252afh:
## Context

Currently the ECS tasks can be deployed and fail to start, but we don't see that in the deployment

Also, the lambda is currently set by env tag which is always behind the current deployment which could cause migration issues

- Add `wait_for_ready_state` to ecs tasks to wait for them to come up
- Changed lambda to use the same image version as the ecs tasks
- Update the auto_apply make command to also update the lambda image

## Things to check


gecBurton (approved):
LGTM

PR #915, opened 2024-08-05
Feature/clarify django status

rachaelcodes:
## Context
After a number of changes to the FileStatus in both the Django app and the core API, the engineers discussed simplifying our approach so that the two types of status only contain information that is relevant for each application and to reduce the implicit dependency on each other.

Only use these file statuses in Django: `processing`, `complete`, `errored`, `deleted`, and move the core_api -> django FileStatus mapping to `models.py`. Update tests as required. 

I have also set the file status to `errored` where there are problems communicating with the core-api, to avoid repeated broken requests. This may require us to offer the option to re-ingest errored files at another time.

## Guidance to review
I've two questions about the approach:
* Should we hold off setting the file status to `errored` in the case of non-404 errors from the core-api?
* Is there a better approach to avoiding unexpected behaviours from new core-api statuses than just the comment I've added to `redbox-core/redbox/models/file.py` ?

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-576

## Things to check


django_app/redbox_app/redbox_core/views/document_views.py
```diff
@@ -188,20 +188,18 @@ def file_status_api_view(request: HttpRequest) -> JsonResponse:
...
     except RequestException as ex:
-        logger.exception("File object information from core not found - file does not exist %s.", file_id, exc_info=ex)
-        if not file.status:
-            file.status = StatusEnum.unknown.label
-            file.save()
-        return JsonResponse({"status": file.status})
-    file.status = core_file_status_response.processing_status or StatusEnum.unknown.name
-    file.save()
+        logger.exception("File object information from core not found - for file %s.", file_id, exc_info=ex)
+        file.status = StatusEnum.errored
```
rachaelcodes, line 201: Should we hold off setting the file status to errored in the case of non-404 errors from the core-api?

redbox-core/redbox/models/file.py
```diff
@@ -13,6 +13,18 @@
...
+
+    Note: The Django app interprets these as:
+    "processing" -> "processing"
+    "complete" -> "complete"
+    "failed" -> "errored"
+    anything else -> "processing"
+
+    If you add any other fail state options, the Django app will need to be updated.
+    django_app/redbox_app/redbox_core/models.py File.update_status_from_core
+    """
```
rachaelcodes, line 26: Is there a more elegant approach here than just pointing out the risk?

django_app/redbox_app/redbox_core/client.py
```diff
@@ -59,21 +58,7 @@ class CoreChatResponse:
...
-            "indexing",
-            "complete",
-            "unknown",
-            "errored",
-            "deleted",
-            "processing",
-        ]
-        | None
-    )
+    processing_status: str
```
gecBurton, line 61: :+1:

django_app/redbox_app/redbox_core/management/commands/check_file_status.py
```diff
@@ -42,9 +42,7 @@ class Command(BaseCommand):
     def handle(self, *_args, **_kwargs):
         self.stdout.write(self.style.NOTICE("Checking file status"))
 
-        statuses_to_ignore = [StatusEnum.complete, StatusEnum.deleted, StatusEnum.errored]
-
-        for file in File.objects.exclude(status__in=statuses_to_ignore):
+        for file in File.objects.filter(status=StatusEnum.processing):
```
gecBurton, line 45: this is a lot clearer

django_app/redbox_app/redbox_core/views/document_views.py
gecBurton: in the case where an error might get resolved in the future i guess we might, but is this ever the case?

django_app/tests/views/test_document_views.py
```diff
@@ -200,6 +200,96 @@ def test_remove_nonexistent_doc(alice: User, client: Client):
...
+    # When
+    response = client.get("/file-status/", {"id": uploaded_file.id})
+
+    # Then
+    assert File.objects.get(id=uploaded_file.id).status == "errored"
+    assert response.json()["status"] == "Errored"
+
+
+@pytest.mark.django_db()
+def test_file_status_api_view_file_core_api_error(
```
gecBurton, line 275: nitpick: this could be merged with a parameterised `test_file_status_api_view_file_core_api_error` ?

redbox-core/redbox/models/file.py
gecBurton: I dont think so. not without boiling the ocean anyway :)

gecBurton (changes requested):
LGTM

some nitpicking added which can be ignored.

please can we test the data migration with a test

django_app/tests/test_migrations.py
```diff
@@ -77,3 +77,54 @@ def test_0020_remove_chatmessage_source_files_textchunk_and_more(migrator):
 
     # Cleanup:
     migrator.reset()
+
+
+@pytest.mark.django_db()
+def test_0027_alter_file_status(migrator):
+    # Not using test parametrisation to avoid repeatedly rerunning migration
```
rachaelcodes, line 84: Happy to change this to test parametrisation if more readable

gecBurton: nah not needed. migration tests are long running enough

gecBurton (approved):
LGTM

PR #923, opened 2024-08-06
Show loading message when uploading large files

KevinEtchells:
## Context

When uploading large files, the UI doesn’t give any feedback to show that the file is uploading. This could cause confusion and/or lead to duplicate uploads.

* Add an uploading message (the same one as used for messages)

## Guidance to review

* Upload a large PDF (I have a 100MB one I can pass on if that's helpful)

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-618

## Things to check


gecBurton (approved):
LGTM

PR #930, opened 2024-08-07
Feature/stop streaming button

KevinEtchells:
## Context

Being able to stop a stream has been a highly requested feature. We have this functionality as an escape key, but many users won’t be aware of or think of this - a button would be better.

I've tried two approaches: (A) having a button in the message container, and (B) ChatGPT-style where the send button turns into a stop button. I've opted for B in the end, as A was a bit clunky when scrolling, and there is already a design for B. A is a separate commit on this branch should we ever wish to go back to it.

Make the send button turn into a stop button while streaming is happening.

## Guidance to review

* Ask Redbox to send a long response and click the stop button part way through
* Also try pressing the escape key to check that still works

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-619

## Things to check


gecBurton (approved):
works for me

PR #947, opened 2024-08-12
Feature/simplify consumers

gecBurton:
## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/consumers.py
+++ b/django_app/redbox_app/redbox_core/consumers.py
@@ -57,7 +57,6 @@ class ChatConsumer(AsyncWebsocketConsumer):
         selected_file_uuids: Sequence[UUID] = [UUID(u) for u in data.get("selectedFiles", [])]
         user: User = self.scope.get("user", None)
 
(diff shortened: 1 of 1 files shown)


django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -121,10 +121,16 @@ async def receive_llm_responses(
     async def handle_documents(
```
rachaelcodes, line 140: Is it worth adding a docstring here? It's not the easiest method to make sense of (particularly `files_and_citations` and an explanation of why we're sending something different to the client than we're returning)

gecBurton: yep, good point

rachaelcodes (approved):
Looks a lot clearer, nice one!

django_app/redbox_app/redbox_core/consumers.py
gecBurton: [b06110f](https://github.com/i-dot-ai/redbox/pull/947/commits/b06110fc73929c7f013e35d569266f9b51ec568e)

PR #959, opened 2024-08-13
Add error message for users in case of 429 errors

rachaelcodes:
## Context
Currently we show users the same error message for all openai API failures. Recently, we've had some rate limiting errors and it could be useful to share this information with users. 

Handles `openai.RateLimitError` exceptions separately from generic exceptions, passing a specific message to the user.

Screenshot 2024-08-13 at 13 26 16

I think the Django error handling code needs some refactoring, but we should get this fix out soon and then work on the refactoring when we have more capacity.

## Guidance to review
Try it out locally (and change core-api/core_api/routes/chat.py to return the rate limit error).

### Questions
* Is there more that we can do, given our current setup, to test this?
* Are we happy with this text?

## Things to check

### DIFF
--- a/core-api/core_api/routes/chat.py
+++ b/core-api/core_api/routes/chat.py
@@ -116,12 +116,14 @@ async def rag_chat_streamed(
             ClientResponse(resource_type="error", data=ErrorDetail(code="rate-limit", message=type(e).__name__)),
             websocket,
         )
+        await websocket.close()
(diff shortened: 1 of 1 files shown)


KevinEtchells (approved):
LGTM. I've added a message on Slack about whether we should also add something on the support page, but that can be a separate PR if needed.

PR #966, opened 2024-08-15
LLM Selector

KevinEtchells:
## Context

This is work-in-progress to add an LLM selector to the UI. 

## Work done so far:

* Added in the selector
* Passing this back to Django on form submit and via websocket (progressive enhancement)

image

## Work still to do

* ORM side - where should this sit / where do we need to store it
* Pass it through to the view
* Handle the form submit and websocket on the Django side

## Other notes

* I've planned for this data structure in the template:
```
{% set llm_options = [
    {"name": "gpt-4o", "default": True, "selected": True},
    {"name": "gpt-4-turbo-2024-04-09"},
    {"name": "gpt-35-turbo-16k"}
] %}
```
`default` because I want to add the words "(default)" next to the default option. So users who don't understand this at least know what option should be selected as standard.
Once the view is passing this in then this hard-coded version can be removed!

* It's being passed back to Django as `llm` both via for form submit and via the web-socket.


gecBurton (approved):
working for me :)

PR #981, opened 2024-08-21
Tracking tokens in state object

jamesrichards4:
## Context

We want to be able to track the token usage of an individual request to Redbox and ultimately per chat/user

Added a new metadata object to the graph state, this currently contains input and output tokens.
Update the token counts on every LLM call

## Guidance to review

This doesn't account for different LLMs, we should probably review this when we start using them. We would certainly need to account for using different LLMs in a single call (ie condense question with 4o-mini and the full context question with 4o) to allow proper cost tracking but we should also maybe record the LLM so as we switch in different releases we have a record?

This metadata doesn't go anywhere yet. Returning it to django could be done in a separate ticket, this work could make django more tolerant of the websocket event types it allows to support this work and future additions

## Things to check

### DIFF
--- a/redbox-core/redbox/models/chain.py
+++ b/redbox-core/redbox/models/chain.py
@@ -180,11 +180,29 @@ class RedboxQuery(BaseModel):
     ai_settings: AISettings = Field(description="User request AI settings", default_factory=AISettings)
 
 
+class RequestMetadata(TypedDict):
+    input_tokens: int
+    output_tokens: int
+
+
+def metadata_reducer(current: RequestMetadata | None, update: RequestMetadata | None):
+    if current is None:
+        return update
+    if update is None:
+        return current
+    if "input_tokens" in update:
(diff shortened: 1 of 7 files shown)


redbox-core/redbox/models/chain.py
```diff
@@ -180,11 +180,29 @@ class RedboxQuery(BaseModel):
...
+    output_tokens: int
+
+
+def metadata_reducer(current: RequestMetadata | None, update: RequestMetadata | None):
+    if current is None:
+        return update
+    if update is None:
+        return current
+    if "input_tokens" in update:
+        input_tokens = current.get("input_tokens", 0) + update["input_tokens"]
```
gecBurton, line 194: ive been bitten by this before where the RequestMetadata is something like `{"input_tokens": None}` and you end up trying to add None to int or similar. do we need something to coerce None to 0?

gecBurton (approved):
nice solution. one nitpick which you can ignore

PR #987, opened 2024-08-23
Add a conditional that checks that document tokens don't exceed n

ghost:
## Context

We're experiencing brownouts where high volume users are triggering rate limits. This allows us to set a hard limit to the amount of document tokens the system will accept.

* Adds a `max_document_tokens` variable to `AISettings`
    * Add to both core and Django, and default to 1m
* Adds `documents_bigger_than_n_conditional()` which routes based on it
* Adds a unit test condition to check it triggers
* Rename `TestData` to `RedboxTestData` as things that start with `test*` [make pytest's ears perk up](https://docs.pytest.org/en/stable/example/pythoncollection.html#changing-naming-conventions) 🐶 

### Proposed graph change

redbox-ntoken-v2

### Implemented graph change

redbox-ntoken-built

## Guidance to review

* Confirm 1m tokens is a sensible default for cutting users off
* Check what I've done with environment variables -- I'm 99% sure I've covered all environments but would appreciate a second eye
* Check my work with Django and migrations, I don't work with Django often
* Check you're happy that this is set in `AISettings` rather than as an environment variable
* Ensure unit tests pass

## Things to check

- [x] [I have run integration tests](https://github.com/i-dot-ai/redbox/actions/runs/10527295574/job/29170136929)


django_app/redbox_app/redbox_core/models.py
```diff
@@ -52,7 +52,9 @@ class ChatBackend(models.TextChoices):
         GPT_4_OMNI = "gpt-4o", _("gpt-4o")
 
     label = models.CharField(max_length=50, unique=True)
-    context_window_size = models.PositiveIntegerField(default=8_000)
+    max_document_tokens = models.PositiveIntegerField(default=1_000_000)
```
gecBurton, line 55: all these new fields need to have null=True and blank=True

django_app/redbox_app/redbox_core/migrations/0037_aisettings_max_document_tokens_and_more.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 5.1 on 2024-08-23 12:25
+
+from django.db import migrations, models
```
gecBurton, line 3: this migration is now out of step with main, pls can you regenerate it

gecBurton (commented):
some Django changes pls. otherwise all looks good

gecBurton (approved):

PR #996, opened 2024-08-28
Feature/django control file directly

gecBurton:
## Context

As an Engineer I want to:
* delete chunks stored in elastic directly
* not create/delete files as this is now unnecessary
* facilitate a move away from core-api

## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox/actions/runs/10667053466


django_app/tests/conftest.py
```diff
@@ -26,6 +28,34 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture()
```
gecBurton, line 31: cut and paste from the now deleted client.py as this is the only place that where it is used

django_app/redbox_app/redbox_core/management/commands/delete_expired_data.py
```diff
@@ -58,16 +56,16 @@ def handle(self, *_args, **_kwargs):
                 )
 
                 try:
-                    core_api.delete_file(file.core_file_uuid, file.user)
+                    file.delete_from_elastic()
                     file.delete_from_s3()
 
-                except RequestException as e:
```
gecBurton, line 64: http-request error is no longer a thing

django_app/redbox_app/redbox_core/models.py
```diff
@@ -295,6 +300,14 @@ def delete_from_s3(self):
         """Manually deletes the file from S3 storage."""
         self.original_file.delete(save=False)
 
+    def delete_from_elastic(self):
+        index = f"{env.elastic_root_index}-chunk"
+        if es_client.indices.exists(index=index):
+            es_client.delete_by_query(
+                index=index,
+                body={"query": {"term": {"metadata.file_name.keyword": self.unique_name}}},
```
jamesrichards4, line 308: Will this delete files with this name for all users? If two users upload the same file and one deletes it, they'll both go?

jamesrichards4 (commented):
There's some potential messiness brewing with the Elastic File datamodel being present in core-api and the django-app. Is your thinking that this is temporary? Once we drop core-api I guess we can clean this up and remove File completely?

gecBurton:
> There's some potential messiness brewing with the Elastic File datamodel being present in core-api and the django-app. Is your thinking that this is temporary? Once we drop core-api I guess we can clean this up and remove File completely?

absolutely. File should be replaced throughput with the globally unique_name/s3-key

jamesrichards4 (approved):

PR #990, opened 2024-08-27
feature/added json-download

gecBurton:
## Context
We want to be able to see useful data quickly, particularly for tracking token spending.

## Changes proposed in this pull request
Adds a new 'download json' option under the User model in Admin, which - thanks to the magic ⚡ of Django Rest Framework Serialisers 🧙 - allows us to see Chats, ChatMessages, token and File usage.

## Guidance to review
To try this locally, go to Django Admin > Users . Click on the select box to select all Users, and choose 'Export Selected' from the dropdown menu. _Does the json look reasonable?_

(Note: this is a different flow from clicking the 'Export' button, which is based on the User model alone and is a default export implementation that allows us to get quick, uncustomised, data)

## Things to check

- [ ] I have added any new ENV vars in all deployed environments
- [x] I have tested any code added or changed
- [ ] I have run integration tests

PR #998, opened 2024-08-29
Tracks tokens in Django

rachaelcodes:
## Context
We need to track message tokens, now that they are being streamed from core-api.

Adds a new `ChatMessageTokenUse` model to track how tokens are used for chat messages.

## Guidance to review
Does it work? What do you think about the data model in Django? 
Should `ChatMessageTokenUse` be exported anywhere in Django Admin (probably a new PR, tbh)

## Things to check


redbox-core/redbox/models/chat.py
```diff
@@ -83,12 +83,17 @@ class ChatResponse(BaseModel):
     route_name: ChatRoute = Field(description="the conversation route taken")
 
 
+class MetadataDetail(BaseModel):
+    input_tokens: dict[str, int] | None = None
+    output_tokens: dict[str, int] | None = None
+
```
rachaelcodes, line 89: This is a near copy of: `redbox-core/redbox/models/chain.py:185`. Should one file import the class from other?

Also, are the None options reasonable? (I think I have seen responses with only one of the two token types listed).

django_app/redbox_app/redbox_core/admin.py
```diff
@@ -74,11 +74,22 @@ class CitationInline(admin.StackedInline):
     extra = 0
 
 
+class ChatMessageTokenUseInline(admin.StackedInline):
```
gecBurton, line 77: love the use of inline :)

django_app/redbox_app/redbox_core/models.py
```diff
@@ -481,3 +481,19 @@ def get_messages_ordered_by_citation_priority(cls, chat_id: uuid.UUID) -> Sequen
...
             )
         )
+
+
+class ChatMessageTokenUse(UUIDPrimaryKeyBase, TimeStampedModel):
+    class UseTypeEnum(models.TextChoices):
+        INPUT = "input", _("input")
+        OUTPUT = "output", _("output")
+
+    chat_message = models.ForeignKey(ChatMessage, on_delete=models.CASCADE)
```
gecBurton, line 491: > What do you think about the data model in Django?

I handnt appreciated that there are many token counts per message, but this does make sense

> should ChatMessageTokenUse be exported anywhere in Django Admin (probably a new PR, tbh)

not sure about admin...  but it should definitely be included here: https://github.com/i-dot-ai/redbox/pull/990

redbox-core/redbox/models/chat.py
gecBurton: ah how annoying! unfortunately I think thank laggraph requires us to use TypedDict rather than pydantic so im not sure we can solve this. @wpfl-dbt what do you think?

django_app/redbox_app/redbox_core/models.py
gecBurton: but , yes, another PR for sure

redbox-core/redbox/models/chat.py
rachaelcodes: I'm tempted to just leave the two definitions rather than tie ourselves in knots trying to link them - especially as it's only the one repetition.

django_app/redbox_app/redbox_core/models.py
```diff
@@ -481,3 +481,19 @@ def get_messages_ordered_by_citation_priority(cls, chat_id: uuid.UUID) -> Sequen
...
+    use_type = models.CharField(
+        max_length=10, choices=UseTypeEnum, help_text="input or output tokens", default=UseTypeEnum.INPUT
+    )
+    model_name = models.CharField(max_length=50, null=True, blank=True)
+    tokens = models.PositiveIntegerField(null=True, blank=True)
```
gecBurton, line 496: just to be super annoying... i think this should be `token_count`

django_app/redbox_app/redbox_core/models.py
```diff
@@ -481,3 +481,19 @@ def get_messages_ordered_by_citation_priority(cls, chat_id: uuid.UUID) -> Sequen
...
+class ChatMessageTokenUse(UUIDPrimaryKeyBase, TimeStampedModel):
+    class UseTypeEnum(models.TextChoices):
+        INPUT = "input", _("input")
+        OUTPUT = "output", _("output")
+
+    chat_message = models.ForeignKey(ChatMessage, on_delete=models.CASCADE)
+    use_type = models.CharField(
+        max_length=10, choices=UseTypeEnum, help_text="input or output tokens", default=UseTypeEnum.INPUT
+    )
+    model_name = models.CharField(max_length=50, null=True, blank=True)
```
gecBurton, line 495: in principle i think this should be https://github.com/i-dot-ai/redbox/blob/fb949dbe5b67b64be69955c6e0a52740b9227525/django_app/redbox_app/redbox_core/models.py#L49 but lets not worry about that now

django_app/redbox_app/redbox_core/models.py
rachaelcodes: > I handnt appreciated that there are many token counts per message, but this does make sense

it gives us a bit of leeway if we decide to go down a path of using different models per message, which James thought was possible.

gecBurton (approved):
great stuff - as always!

some nitpicking added, none of it is important. :shipit:

redbox-core/redbox/models/chat.py
gecBurton: yep, happy with this.

PR #999, opened 2024-08-29
Added env integration layer to django and db in docker-compose

jamesrichards4:
## Context

We want to allow clean usage of env files in different places. We're currently not using env integration for django in docker-compose

Use .env.integration and .env in django and db in docker-compose rather than just .env. This means .env can leave any hosts out to allow also using it in unit tests and vscode launch profiles without issue.

## Guidance to review

This shouldn't affect anyone who isn't using vscode launch profiles and docker compose. Any questions about env let me know, could be worth a write up on how we're using them across the different ways to run

## Things to check

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,10 @@ services:
     networks:
       - redbox-app-network
     env_file:
-      - .env
+      - path: .env.integration
+        required: false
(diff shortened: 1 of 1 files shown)


gecBurton (approved):
good idea

PR #1003, opened 2024-08-29
feature/added parameters for django-q

gecBurton:
## Context

As an Engineer I want to:
* increase the timeout for the unstructured tasks
* set a max_retry 
* set these parameters from env vars so that we dont have to rebuild every time we want to change a paremter

Large files are more likely to succeed given enough time, but we dont indefinitely retry for file that are too large for the timeout

## Things to check

### DIFF
--- a/django_app/redbox_app/settings.py
+++ b/django_app/redbox_app/settings.py
@@ -337,8 +337,9 @@ REDBOX_VERSION = os.environ.get("REDBOX_VERSION", "not set")
 
 Q_CLUSTER = {
     "name": "redbox_django",
-    "timeout": 120,
-    "retry": 600,
+    "timeout": os.environ.get("Q_TIMEOUT", 300),
+    "retry": os.environ.get("Q_RETRY", 900),
+    "max_attempts": os.environ.get("Q_MAX_ATTEMPTS", 3),
     "catch_up": False,
     "orm": "default",
 }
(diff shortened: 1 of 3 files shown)


django_app/redbox_app/settings.py
```diff
@@ -337,8 +337,9 @@
 
 Q_CLUSTER = {
     "name": "redbox_django",
-    "timeout": 120,
-    "retry": 600,
+    "timeout": os.environ.get("Q_TIMEOUT", 300),
+    "retry": os.environ.get("Q_RETRY", 900),
+    "max_attempts": os.environ.get("Q_MAX_ATTEMPTS", 3),
```
gecBurton, line 342: this is the key change

infrastructure/aws/variables.tf
```diff
@@ -484,3 +484,21 @@ variable "worker_ingest_largest_chunk_overlap" {
...
+
+variable "django_queue_retry" {
+  type        = number
+  default     = 300
+  description = "How long to wait between retrying unstructured task"
+}
+
+variable "django_queue_max_attempts" {
+  type        = number
+  default     = 300
```
gecBurton, line 502:
```suggestion
  default     = 3
```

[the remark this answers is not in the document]
gecBurton: strongly agree!

PR #1006, opened 2024-08-30
feature/user-admin-is_developer

gecBurton:
## Context

As an Engineer I want to to be able to edit the User.is_developer field so that i can exclude i.AI engineers from UR analysis

## Things to check


django_app/redbox_app/redbox_core/admin.py
```diff
@@ -41,16 +41,14 @@ def export_as_json(self, request, queryset: QuerySet):  # noqa:ARG002
         "is_staff",
         "last_login",
         "ai_settings",
+        "is_developer",
     ]
     list_display = [
         "email",
-        "name",
```
gecBurton, line 47: this works best with a small number of fields

PR #1008, opened 2024-09-02
Making the token count check do both max_tokens and context_window ch…

jamesrichards4:
We were previously only check max_tokens when the request total tokens exceeded the context window. To avoid confusing situations with the interaction of these env vars both are now always checked so that an error is thrown when the total tokens exceeds the max regardless of whether it's more or less than the context window

A new decision node returns a request handler name from 'max_exceeded'/'context_exceeded'/'pass' rather than two boolean checks in serial. This allows moving to the correct next node in all situationss

## Guidance to review

Ignore the reformatting in django-app. Must have missed a Ruff pass on a previous PR

## Things to check

### DIFF
--- a/redbox-core/redbox/graph/edges.py
+++ b/redbox-core/redbox/graph/edges.py
@@ -1,5 +1,6 @@
 import logging
 import re
+from typing import Literal
 
 from langchain_core.runnables import Runnable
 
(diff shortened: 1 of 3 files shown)


redbox-core/redbox/graph/edges.py
```diff
@@ -24,6 +25,30 @@ def calculate_token_budget(state: RedboxState, system_prompt: str, question_prom
...
+    ) -> Literal["max_exceeded", "context_exceeded", "pass"]:
+        system_prompt, question_prompt = get_prompts(state, prompt_set)
+        token_budget_remaining_in_context = calculate_token_budget(state, system_prompt, question_prompt)
+        max_tokens_allowed = state["request"].ai_settings.max_document_tokens
+
+        total_tokens = sum(d.metadata["token_count"] for d in flatten_document_state(state["documents"]))
+
+        if total_tokens > max_tokens_allowed:
+            return "max_exceeded"
+        elif total_tokens > token_budget_remaining_in_context:
```
gecBurton, line 44:
```suggestion
        if total_tokens > token_budget_remaining_in_context:
```

redbox-core/redbox/graph/edges.py
```diff
@@ -24,6 +25,30 @@ def calculate_token_budget(state: RedboxState, system_prompt: str, question_prom
...
+            return "context_exceeded"
+        else:
```
gecBurton, line 46:
```suggestion
```

gecBurton (approved):
some nitpicking added, feel free to ignore

PR #1012, opened 2024-09-03
feature/added  aws bedrock

gecBurton:
## Context

As a PM i want to be able to choose freely from Azure and AWS so that I can offer our users a wide choice of models

I have included both the LLM selection (via django/ai-settings) and the Embedding selection (via global settings/env-vars)

Note that: I have take the opportunity to rationalise the way we specify the embedding-backend with a single variable `embedding_backend` (similar to `chat_backend`) rather than `embedding_backend` and one of:
* `azure_embedding_model`
* `embedding_openai_model`
* `embedding_aws_model`

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-653

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox/actions/runs/10686002852


redbox-core/redbox/chains/components.py
```diff
@@ -87,7 +95,7 @@ def get_azure_embeddings(env: Settings):
         api_key=convert_to_secret_str(env.embedding_openai_api_key),
         azure_endpoint=env.embedding_azure_openai_endpoint,
         api_version=env.azure_api_version_embeddings,
-        model=env.azure_embedding_model,
+        model=env.embedding_backend,
```
gecBurton, line 98: this now follows the same pattern as `get_llm_model`

infrastructure/aws/variables.tf
```diff
@@ -502,3 +502,10 @@ variable "django_queue_max_attempts" {
   default     = 1
   description = "How many attempts to run unstructured task"
 }
+
+variable "embedding_backend" {
+  type        = string
+  default     = "text-embedding-ada-002"
```
jamesrichards4, line 508: just to be safe should we set this to our current model? text-embedding-3-large?

jamesrichards4 (approved):
One small note on a TF var but feel free to ignore. Only other thing, which doesn't matter for now, is that we may want to think about how to allow a deployment to use a different model name for a given model provider without us having to update the settings. The backend, model separation allowed this, this new style is neater but perhaps more limiting across deployments?

infrastructure/aws/variables.tf
gecBurton: agreed

gecBurton, line 508:
```suggestion
  default     = "text-embedding-3-large"
```

PR #1017, opened 2024-09-04
Add a reporting dash directly into Django admin

ghost:
## Context

We want to be able to report on what's going on with users. Instead of building complex pipelines, why not do it directly in Django?

This is a spike PR so I can show and discuss my work with others. I did it in my own time and am completely happy for it to be dropped if it's not useful.

Because this work is intended to enable data scientists rather than Django developers, the plotly app has some extra structure so it can be run outside of Django as well as inside -- the only dependency is `docker compose up db -d`.

Out of scope is actually making the report useful or beautiful: it just needs to work.

> [!WARNING]  
> A migration is needed to install [django-plotly-dash](https://django-plotly-dash.readthedocs.io/en/latest/installation.html), even though no scripts have been generated with `makemigrations`. How do I handle this if we merge/deploy? I don't want to break anything.

* Add [django-plotly-dash](https://pypi.org/project/django-plotly-dash/) to render plotly apps
* Add [django-adminplus](https://pypi.org/project/django-adminplus/) to allow custom views in admin
* Add a barebones demo app for others to edit and extend
* I'm really sorry @gecBurton but `ImportError: Plotly express requires pandas to be installed.` 😬 

Addition of custom view in admin:

Screenshot 2024-09-04 at 23 23 20

Basic report of messages per day:

Screenshot 2024-09-04 at 23 23 37

## Guidance to review

* Spin it up in Django and take a look!
* Run the app in its script alone and take a look... again!
* Ensure modifications to templating engine are fine
* Ensure updates to CSP and xframe in middleware are appropriate
* Confirm you're happy that the extra complexity in the app to enable standalone running is worth it

## Things to check

- [x] [I have run integration tests](https://github.com/i-dot-ai/redbox/actions/runs/10716549058)

### DIFF
--- a/django_app/redbox_app/redbox_core/dash_apps/report_app.py
+++ b/django_app/redbox_app/redbox_core/dash_apps/report_app.py
@@ -6,6 +6,11 @@ from django.db.models import Count
 from django.db.models.functions import TruncDate
 from plotly.graph_objects import Figure
 
+# ----------------------------------#
+# App setup and conditional imports #
+# ----------------------------------#
+
+
 if __name__ == "__main__":
     # Enables standalone running for developers outside of full Django app
     import os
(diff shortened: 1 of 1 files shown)


django_app/redbox_app/redbox_core/dash_apps/report_app.py
```diff
@@ -0,0 +1,83 @@
...
+        .annotate(count=Count("id"))
+        .order_by("day")
+    )
+
+    data = (
+        pd.DataFrame.from_records(queryset.values())
+        .filter(["created_at", "count"])
+        .assign(created_at=lambda x: pd.to_datetime(x["created_at"].dt.strftime("%Y-%m-%d")))
+        .rename(columns={"created_at": "Date", "count": "Count"})
+        .groupby(["Date"], as_index=False)
```
gecBurton, line 75: strong feelings about this - get PG to do the hard work not python - this will be much more efficient

gecBurton (approved):
lovely

PR #1028, opened 2024-09-09
Use versioned terraform modules (alb, waf, ecs)

RyanWhite25:
## Context

Use versioned terraform modules from git (alb, waf, ecs)

## Things to check


gecBurton (approved):
LGTM

PR #1029, opened 2024-09-09
Fixed settings to unstructured to make largest chunks the proper size

jamesrichards4:
## Context

Our 'largest' chunk resolution had no chunking and so were very small. This returns them to being the proper size

Use a single configurable document loader for all chunk resolutions. This uses chunk_by_title and sets the min and max chunk sizes to ensure we get the configured sizes.

## Guidance to review

We probably want to add a new env var for largest_min_chunk_size at some point and there are improvements coming to the actual numbers. This PR just addresses the regression and a refactor which makes sense with this new config

## Things to check


gecBurton (approved):
LGTM

PR #1040, opened 2024-09-11
Filter out /admin from Sentry

KevinEtchells:
## Context

We (probably) don't need Sentry to run on the Django Admin pages. We're getting a lot of errors around the use of jQuery and inline styles that would be a disproportionate amount of work to fix. This will also make managing "real" errors easier to discover and manage in Sentry.

* Filter out all admin pages

## Notes

* This hasn't been properly tested yet - we can monitor this once deployed
* Does this sound like a sensible thing to do?
* Does the code look sensible?

## Things to check

### DIFF
--- a/django_app/redbox_app/settings.py
+++ b/django_app/redbox_app/settings.py
@@ -2,6 +2,7 @@
 import os
 import socket
 from pathlib import Path
+from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


gecBurton (approved):
LGTM

PR #1050, opened 2024-09-13
Added django launch config and npm build task

jamesrichards4:
## Context

As an AI Engineer I want to be able to run the django-app locally, outside of docker, to allow me to use the vscode debugger. I also don't want to have to think too much about frontend assets and want it to 'just work'.

Having the ability to debug should make things easier for anyone using vscode though, not just AI engineering work.

Added a launch config for the django app. This uses the root level .env file and overrides hostnames to be localhost. This means we only need one .env and it can contain the hostnames used in docker deployments. The app can be launched using the run & debug tab in the sidebar as normal.

Added a task for npm run build. If a user has installed and configured npm this will create the frontend assets for them. The easiest way to run the task is to hit f1 and search 'run task', it will then prompt you to run 'Build Frontend'

## Guidance to review

This simplifies the env file setup I think? Is there a better way to do this?

## Things to check


KevinEtchells (approved):
Looks good!

PR #1055, opened 2024-09-18
Working in unit tests

jamesrichards4:
## Context

We want to report what Redbox is doing to users. Choices made, details about which files were read etc

Create a new activity event type and a node to emit these as needed in the graph. These events can be emitted from inside another node as well though.

## Guidance to review

Do we want these event emitting nodes? Is there a nicer way to pull these dispatch calls into nodes?
Are we ok with the structure of the events, is just a str better? Does having a BaseModel gives us flexibility in future?

## Things to check

### DIFF
--- a/django_app/.vscode/launch.json
+++ b/django_app/.vscode/launch.json
@@ -14,7 +14,8 @@
             "env": {
                 "MINIO_HOST": "localhost",
                 "POSTGRES_HOST": "localhost",
-                "UNSTRUCTURED_HOST": "localhost"
+                "UNSTRUCTURED_HOST": "localhost",
+                "ELASTIC__HOST": "localhost"
             }
         }
         },
--- a/redbox-core/redbox/app.py
+++ b/redbox-core/redbox/app.py
@@ -39,6 +39,7 @@ class Redbox:
         route_name_callback=_default_callback,
         documents_callback=_default_callback,
         metadata_tokens_callback=_default_callback,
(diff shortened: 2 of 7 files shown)


jamesrichards4:
I think activity events are only for comms back to the user. If we use them for what is essentially logging I think we might end up tangled up somehow. I ended up with a new node here to avoid emitting an event on every node execution. Basically if we put the activity events in the nodes then the user gets something like:
Set route to chat/documents
Called Stuff
Set route to search
Called Stuff
Called Stuff

This is developer logging, so we'd need some way to only create events when needed. One thing I considered was making it possible to pass in events to be triggered for each pattern so we avoid additional nodes? I had hoped with the nodes they would just be 'off to the side', i.e. an additional edge that goes to an event node is added when we want to do that. The problem is conditional edges don't allow that, you can only go to one node for each path.

Maybe the paths here are:
* Add 'audience' or similar to the events so we emit only activity events with audience: user for use in the activity log
* Make nodes optionally emit events, to avoid event nodes
* Use path functions rather than maps to allow passing execution to the next processing node and an event node without messing up the graph structure

I'm kind of leaning the first one now I think about it. We could piggyback on this to provide our event logging to elastic by having a dev audience?

Thoughts? Might be worth a chat

lmwilkigov (approved):
Gosh this langgraph stuff makes this code so elegant. Love it!

PR #1060, opened 2024-09-23
Propagate retriever signal to neighbours and join retrieved chunks

ghost:
## Context

As part of the [agentic RAG spike](https://github.com/i-dot-ai/redbox/blob/spike/rag/notebooks/rag2.ipynb) I built a retriever that:

* Propagates signal from hits to chunks that are nearby in the document
* Orders the retrieved chunks to recreate blocks of text from the original document while still honouring score

This PR is to pull these ideas into the existing RAG system.

It also lays some groundwork for bringing this functionality into an agentic RAG system.

* Refactors `ParameterisedElasticsearchRetriever` so it's now an entirely custom retriever that we build
* Refactors and extends `get_some` into `build_document_query`, which with `merge_documents`, can handle the propagation of signal to nearby chunks
* Adds `sort_documents`, which will reorder results into blocks of `group_uuid`s of consecutive indices, where the block is ordered by its highest-scoring member
* Parameterises `build_retriever_pattern` so different ways of ordering documents in the state are possible
* Adds the mocking of document scores and indices in unit tests
* Add new parameters that control signal propagation to `AISettings`
* Fixes retriever bug, where RAG was reading `state.request.question` instead of `state.text` as intended
    * This would have meant condense question was basically pointless

## Guidance to review

* Check my unit tests -- this adds quite a few new low-level functions that have some fiddly logic. They need to be right
* Confirm `AISettings` environment variables are updated in the right places. I've given defaults specifically to avoid Terraform and `.env` hell

## Relevant links

* [Custom retriever](https://python.langchain.com/docs/how_to/custom_retriever/)

## Things to check

- [x] [I have run integration tests](https://github.com/i-dot-ai/redbox/actions/runs/11018389147)

### DIFF
--- a/redbox-core/redbox/retriever/queries.py
+++ b/redbox-core/redbox/retriever/queries.py
@@ -92,18 +92,11 @@ def build_document_query(
     permitted_files: list[str],
     selected_files: list[str] | None = None,
     chunk_resolution: ChunkResolution | None = None,
-    adjacent: list[Document] | None = None,
 ) -> dict[str, Any]:
     """Builds a an Elasticsearch query that will return documents when called.
 
     Searches the document:
         * Text, as a keyword and similarity
-        * Name
-        * Description
-        * Keywords
-
-    If given a list of documents in adjacent, will boost adjacent documents by
-    the score of the documents it's been given, scaled.
     """
     # If nothing is selected, consider all permitted files selected
     if not selected_files:
@@ -115,7 +108,7 @@ def build_document_query(
         chunk_resolution=chunk_resolution,
     )
 
-    query_elastic = {
+    return {
         "size": ai_settings.rag_k,
         "query": {
             "bool": {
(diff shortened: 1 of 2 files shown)


redbox-core/tests/test_transform.py
```diff
@@ -183,3 +194,132 @@ def test_to_request_metadata(output: dict, expected: RequestMetadata):
...
+        assert isinstance(group_docs, dict)
+        assert len(group_docs) == n_per_group
+
+        for doc in group_docs.values():
+            assert doc.metadata["uuid"] in group_docs
+            assert group_docs[doc.metadata["uuid"]] == doc
+
+
+def test_merge_documents():
+    docs_1 = list(
```
saisakul, line 266: Can you add docstring to describe what you are testing - the 2 scenarios?

saisakul (approved):

redbox-core/redbox/retriever/queries.py
```diff
@@ -117,18 +144,64 @@ def get_some(
...
+    def scale_score(score: float, old_min: float, old_max: float, new_min=1.1, new_max: float = 2.0):
+        """Rescales an Elasticsearch score.
+
+        Intended to turn the score into a multiplier to weight a Gauss function.
+
+        If the old range is zero or undefined, returns new_min.
+        """
+        if old_max == old_min:
+            return new_min
+
```
lmwilkigov, line 156: Should this be nested in such a big function?

redbox-core/redbox/retriever/queries.py
```diff
@@ -68,47 +68,74 @@ def get_all(
...
+) -> dict[str, Any]:
+    """Builds a an Elasticsearch query that will return documents when called.
+
+    Searches the document:
+        * Text, as a keyword and similarity
+        * Name
+        * Description
+        * Keywords
+
+    If given a list of documents in adjacent, will boost adjacent documents by
```
lmwilkigov, line 105: The code for this func looks good but I think the docstring could do with a bit more narrative to describe what the gaussian adjacency code adds to the function.

redbox-core/redbox/retriever/retrievers.py
```diff
@@ -66,36 +75,56 @@ def _filter_by_elbow(docs: list[Document]) -> list[Document]:
...
+            ai_settings=ai_settings,
+            adjacent=initial_documents,
+        )
+        adjacent_boosted = query_to_documents(
+            es_client=self.es_client, index_name=self.index_name, query=with_adjacent_query
+        )
+
+        # Merge, sort, return
+        merged_documents = merge_documents(initial=initial_documents, adjacent=adjacent_boosted)
+        return sort_documents(documents=merged_documents)
```
lmwilkigov, line 128: This is very elegant 🙌

lmwilkigov (changes requested):

redbox-core/redbox/retriever/queries.py
lmwilkigov: Makes the code a bit hard to follow imo as it is a huge stretch of one level indention

lmwilkigov: I'll approve as deffo just a style comment. Substance looks great

lmwilkigov (approved):

PR #1071, opened 2024-09-27
Add Elastic alias for cleaner ingestion

rachaelcodes:
## Context
When ingesting our files, we should be making sure we remove old artifacts. The standard ElasticSearch approach is to use an 'alias' rather than 'index' as a reference, which allows us to update the index 'behind the scenes' without affecting users. I've started updating the file ingestion to reflect this.

* New django management commands to create and apply the new alias (via our q2 queuing system)
* A new jupyter notebook to simplify dev testing (I will remove this and the new jupyter dependency in a later commit)
* Fix references to elastic to make use of the alias rather than the index
* Update automated tests
* Check/update integration tests
* Move the jupyter notebook to `notebooks` for reference and remove the jupyter poetry dependency

## Guidance to review
To try this out locally, try out the commands as a Scheduled Command from the local django admin in the order listed in the jupyter notebook: 
* add_es_alias
* reingest_files
* change_es_aliased_index
* delete_es_indices

You can check the current list of indices via [http://localhost:9200/_cat/indices](http://localhost:9200/_cat/indices) and the aliased index via [http://localhost:9200/_alias/redbox-data-{environment}-chunk-current](http://localhost:9200/_alias/redbox-data-integration-chunk-current)

Alternatively, you can run the notebook from the `/redbox/django_app` directory after running `poetry install` and `poetry run jupyter notebook`

## Things to check


django_app/redbox_app/redbox_core/management/commands/add_es_alias.py
```diff
@@ -0,0 +1,18 @@
...
+
+from redbox.models import Settings
+
+env = Settings()
+
+es_client = env.elasticsearch_client()
+
+
+class Command(BaseCommand):
+    help = """This is a one-off command to add an ElasticSearch alias to the existing chunks index."""
```
gecBurton, line 11: does this belong in the settings, or even in `def env.elasticsearch_client`?

django_app/redbox_app/redbox_core/management/commands/change_es_aliased_index.py
```diff
@@ -0,0 +1,39 @@
...
+    def add_arguments(self, parser):
+        default_alias = f"{env.elastic_root_index}-chunk-current"
+
+        parser.add_argument("new_index", nargs="?", type=str)
+        parser.add_argument("alias", nargs="?", type=str, default=default_alias)
+
+    def handle(self, *_args, **kwargs):
+        try:
+            response = es_client.indices.get_alias(name=kwargs["alias"])
+            indices_to_remove = list(response.keys())
```
gecBurton, line 32:
```suggestion
            indices_to_remove = list(response)
```
nitpicking

gecBurton (approved):
this is working me for me in dev. would be good to merge some of these commands into one as you suggest (but in a seperate PR)

PR #1072, opened 2024-09-30
feat: integrate google analytics where uktrade is the repo owner

nboyse:
## Context

As part of DBT's ongoing task to contribute back to i.AI's redbox repo, we are using google analytics in the circumstance that the repo owner setting is set to uktrade.

### DIFF
--- a/django_app/redbox_app/jinja2.py
+++ b/django_app/redbox_app/jinja2.py
@@ -95,6 +95,9 @@ def environment(**options):
             "to_user_timezone": to_user_timezone,
             "environment": settings.ENVIRONMENT.value,
             "security": settings.MAX_SECURITY_CLASSIFICATION.value,
+            "repo_owner": settings.REPO_OWNER,
+            "analytics_tag": settings.ANALYTICS_TAG,
+            "analytics_link": settings.ANALYTICS_LINK,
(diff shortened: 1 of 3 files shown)


django_app/redbox_app/jinja2.py
```diff
@@ -95,6 +95,9 @@ def environment(**options):
             "to_user_timezone": to_user_timezone,
             "environment": settings.ENVIRONMENT.value,
             "security": settings.MAX_SECURITY_CLASSIFICATION.value,
+            "repo_owner": settings.REPO_OWNER,
```
252afh, line 98: I'd prefer we used waffle as this is our standards for feature flagging
https://waffle.readthedocs.io/en/stable/

KevinEtchells:
Hi @nboyse, in general this looks fine to me. Just one request:
Would you mind adding `google_` or `GOOGLE_` to all of these? Just thinking of avoiding any confusion if we end up doing something similar with our Plausible Analytics. I appreciate we're hardcoding these for now, but that likely won't always be the case. There's also talk of us adding PostHog at some point, so it's going to get crowded!

KevinEtchells (approved):
LGTM!

PR #1073, opened 2024-10-02
Add an agentic RAG process behind the @gadget keyword

wpfl-dbt:
## Context

It's time to implement all the building blocks of the [agentic RAG spike](https://github.com/i-dot-ai/redbox/blob/spike/rag/notebooks/rag2.ipynb). We want to dogfood and refine the performance of the agent before considering rolling it out. This adds it behind the `@gadget` keyword.

Having dogfooded this a little there's definitely some rough edges but this PR is big enough as it is.

### Known issues

* Unhandled: the agent runs so many searches it exceeds the content window
* The agent takes a relatively long time to deal with impossible questions

The structure of the new route:

agentic4

### Core changes

* Adds an agentic RAG subgraph and accesses it with the `@gadget` keyword
* Adds a send and conditional for use with tools
* Adds the potential to use tools to all core patterns, though only `build_stuff_pattern` is tested
* Adds the count of steps left before the recursion limit is hit to the state
* Updates `Redbox.run` to handle documents returned from tools as sources
* Updates `build_chat_prompt_from_messages_runnable` to allow the insertion of the state's tool calls into the prompt as per OpenAI's standard, in order to show the queries the LLM has already tried
* Updates `build_tool_pattern` to deal with the realities of working with tools, particularly `InjectedState` 😬 
* Moved prompts used by both Django and core to a common directory to import from

### Unit testing and bugfixes

* Adds a `GenericFakeChatModelWithTools` class to enable mocking a tool-calling LLM
* Adds tests for the `documents_callback` of the app

## Guidance to review

* Check you're happy with my unit tests
* Confirm you're happy with the known issues being out of scope for this PR
* I decided to instantiate tools at the top level of the app to avoid passing `Settings()` in. Does that work for you?
* If the tool errors for any reason, I currently mark it as "called". I don't think this is right in the long run. Is this okay for now?
    * Alternative 1: `None` it and remove it
    * Alternative 2: `ToolStateEntry.called` is no longer a bool, and we allow `true`, `false` and `failed` as options (or similar)

## Relevant links

* [Agentic RAG spike](https://github.com/i-dot-ai/redbox/blob/spike/rag/notebooks/rag2.ipynb)
* [How to return state before hitting recursion limit](https://langchain-ai.github.io/langgraph/how-tos/return-when-recursion-limit-hits/)

## Things to check

- [x] [I have run integration tests](https://github.com/i-dot-ai/redbox/actions/runs/11234600830)


redbox-core/redbox/api/format.py
```diff
@@ -32,3 +35,26 @@ def reduce_chunks_by_tokens(chunks: list[Document] | None, chunk: Document, max_
...
+
+
+def format_toolstate(toolstate: ToolState | None) -> str:
+    """Takes a toolstate and transforms it into a structure familiar to an LLM."""
+    if not toolstate:
+        return ""
+
+    formatted_calls: list[str] = []
+
+    for _, call_info in toolstate.items():
```
gecBurton, line 47:
```suggestion
    for call_info in toolstate.values():
```
nitpicking

redbox-core/redbox/app.py
```diff
@@ -22,15 +34,40 @@ def __init__(
...
+
+        search_documents = build_search_documents_tool(
+            es_client=_env.elasticsearch_client(),
+            index_name=f"{_env.elastic_root_index}-chunk",
+            embedding_model=_embedding_model,
+            embedding_field_name=_env.embedding_document_field_name,
+            chunk_resolution=ChunkResolution.normal,
+        )
+
+        tools: dict[str, StructuredTool] = {
```
gecBurton, line 60: just for my understanding - here we have a dict of tools which right now includes just one tool: doc-search, but could be extended to wikipedia, lex, etc?

redbox-core/redbox/app.py
```diff
@@ -42,7 +79,7 @@ async def run(
         activity_event_callback=_default_callback,
     ) -> RedboxState:
         final_state = None
-        async for event in self.graph.astream_events(input, version="v2"):
+        async for event in self.graph.astream_events(input=input, version="v2", config={"recursion_limit": 50}):
```
gecBurton, line 82: just thinking out loud, no need to do anything: 50 seems pretty huge? (in a separate PR lets move this into the AISettings)

redbox-core/redbox/graph/edges.py
```diff
@@ -95,3 +94,29 @@ def documents_selected_conditional(state: RedboxState) -> bool:
...
+        for tool_call in state["tool_calls"].values():
+            if tool_call["tool"]["name"] in tools and not tool_call["called"]:
+                return True
+        return False
+
+    return _tools_selected_conditional
+
+
+def build_strings_end_text_conditional(*strings: str) -> Runnable:
+    """Given a list of strings, returns the string if state["text"] ends with it."""
```
gecBurton, line 112: this doesnt quite do what the comment says it does. why not use `endswith`?

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -51,12 +51,19 @@ def build_retrieve_pattern(
...
 ) -> Runnable[RedboxState, dict[str, Any]]:
-    """Returns a Runnable that uses state["request"] to set state["text"]."""
+    """Returns a Runnable that uses the state to set state["text"].
+
+    If tools are supplied, can also set state["tool_calls"].
+    """
 
     def _chat(state: RedboxState) -> dict[str, Any]:
         llm = get_chat_llm(Settings(), state["request"].ai_settings)
+        if tools:
```
gecBurton, line 64: thinking out loud again: we are repeating this pattern 3 times, is there an easier way? no need to do anything this can be re-visitied later

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -250,20 +270,45 @@ def build_tool_pattern(tools=list[StructuredTool]) -> Callable[[RedboxState], di
     def _tool(state: RedboxState) -> dict[str, Any]:
         state_updates: list[dict] = []
 
-        for tool_id, tool_call_dict in state.get("tool_calls", {}).items():
+        tool_calls = state.get("tool_calls", {})
+        if not tool_calls:
+            log.warning("No tool calls found in state")
+            return None
```
gecBurton, line 276:
```suggestion
            return {}
```
or change annotation?

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -250,20 +270,45 @@ def build_tool_pattern(tools=list[StructuredTool]) -> Callable[[RedboxState], di
...
+
+                    # Invoke the tool
+                    result_state_update = tool.invoke(args) or {}
+                    tool_called_state_update = {"tool_calls": {tool_id: {"called": True, "tool": tool_call}}}
+                    state_updates.append(result_state_update | tool_called_state_update)
+
+                except Exception as e:
+                    state_updates.append({"tool_calls": {tool_id: {"called": True, "tool": tool_call}}})
+                    log.warning(f"Error invoking tool {tool_call['name']}: {e} \n")
+                    return None
```
gecBurton, line 304:
```suggestion
                    return {}
```

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -250,20 +270,45 @@ def build_tool_pattern(tools=list[StructuredTool]) -> Callable[[RedboxState], di
...
                 tool = tools_by_name[tool_call["name"]]
 
-                result_state_update = tool.invoke(tool_call["args"])
-                tool_called_state_update = {"tool_calls": {tool_id: {"called": True}}}
+                if tool is None:
+                    log.warning(f"Tool {tool_call['name']} not found")
+                    continue
 
-                state_updates.append(result_state_update | tool_called_state_update)
+                try:
```
gecBurton, line 288: move this down so that it only wraps `invoke` ?

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -250,20 +270,45 @@ def build_tool_pattern(tools=list[StructuredTool]) -> Callable[[RedboxState], di
...
+                        args["state"] = state
+
+                    log.info(f"Invoking tool {tool_call['name']} with args {args}")
+
+                    # Invoke the tool
+                    result_state_update = tool.invoke(args) or {}
+                    tool_called_state_update = {"tool_calls": {tool_id: {"called": True, "tool": tool_call}}}
+                    state_updates.append(result_state_update | tool_called_state_update)
+
+                except Exception as e:
```
gecBurton, line 301: you would imagine that there is a langchain runtime exception... but there is not https://python.langchain.com/docs/how_to/tools_error/#tryexcept-tool-call 😆

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -275,6 +320,13 @@ def clear_documents_process(state: RedboxState) -> dict[str, Any]:
         return {"documents": {group_id: None for group_id in documents}}
 
 
+def report_sources_process(state: RedboxState) -> None:
+    """A Runnable which reports the documents in the state as sources."""
+    if document_state := state.get("documents"):
+        dispatch_custom_event(RedboxEventType.on_source_report, flatten_document_state(document_state))
+    return None
```
gecBurton, line 327:
```suggestion
    return None
```
this doesnt do anything

redbox-core/redbox/graph/nodes/sends.py
```diff
@@ -41,3 +41,20 @@ def _chunk_send(state: RedboxState) -> list[Send]:
         return [Send(node=target, arg=state) for state in chunk_send_states]
 
     return _chunk_send
+
+
+def build_tool_send(target: str) -> Callable[[RedboxState], list[Send]]:
+    def _tool_send(state: RedboxState) -> list[Send]:
+        if state.get("tool_calls") is None:
```
gecBurton, line 48: this is already built into python and run on line 56?

redbox-core/redbox/app.py
gecBurton: > each step

ahhh, got it, thx

gecBurton (approved):
very cool indeed!

PR #1084, opened 2024-10-08
Feature/control model selection

gecBurton:
## Context

This PR allows us to create and manage LLMs in the django db so that departments can manage their own models

image

A new ChatLLMBackend model has been added that allows a model to be configured according to https://python.langchain.com/api_reference/langchain/chat_models/langchain.chat_models.base.init_chat_model.html

## Things to check

- [x] I have run integration tests https://github.com/i-dot-ai/redbox/actions/runs/11274307158


django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -76,19 +77,16 @@ async def receive(self, text_data=None, bytes_data=None):
...
         selected_file_uuids: Sequence[UUID] = [UUID(u) for u in data.get("selectedFiles", [])]
         user: User = self.scope.get("user")
-        chat_backend = data.get("llm")
         temperature = data.get("temperature")
 
         if session_id := data.get("sessionId"):
             session = await Chat.objects.aget(id=session_id)
-            logger.info("updating: chat_backend=%s -> ai_settings=%s", session.chat_backend, chat_backend)
-            session.chat_backend = chat_backend
+            if llm := data.get("llm"):
```
a-gleeson, line 84: Should we have an info or debug log here to indicate retrieving this from the DB?

django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -76,19 +77,16 @@ async def receive(self, text_data=None, bytes_data=None):
...
+                session.chat_backend = await ChatLLMBackend.objects.aget(name=llm)
             session.temperature = temperature
             await session.asave()
         else:
-            session = await Chat.objects.acreate(
-                name=user_message_text[: settings.CHAT_TITLE_LENGTH], user=user, chat_backend=chat_backend
-            )
+            session = await Chat.objects.acreate(name=user_message_text[: settings.CHAT_TITLE_LENGTH], user=user)
```
a-gleeson, line 89: Ditto above should we log that we are creating a new chat as debug?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -51,22 +51,54 @@ def sanitise_string(string: str | None) -> str | None:
     return string.replace("\x00", "\ufffd") if string else string
 
 
+class ChatLLMBackend(models.Model):
```
a-gleeson, line 54: So this is where your approach slightly differed from my own. I had split the model creation from the available models via the `AiSettings`. The idea of this would be that you could have different AiSettings for different users could mean they have different models registered, stored in the through table `RegisteredModels`. Maybe it is over engineering it at this point, but happy to hear what you think?

a-gleeson: An idea I am also less cold on as this integrate with chat_init_model, is maintaining Embedding models in the same table. This could mean for an application you can possibly in the future have different config with different embedding models and registered indicies. Possibly this is a nice direction for a tests framework, to run comaprisions and record results. Maybe it doesn't live in redbox either and is a seperate thing.

redbox-core/redbox/chains/components.py
```diff
@@ -16,22 +16,16 @@
...
-    elif model.startswith("anthropic."):
-        model_provider = "bedrock"
-    else:
-        raise ValueError("%s not recognised", model)
-
-    logger.info("initialising model=%s model_provider=%s tools=%s", model, model_provider, tools)
-    chat_model = init_chat_model(model=model, model_provider=model_provider)
+def get_chat_llm(model: ChatLLMBackend, tools: list[StructuredTool] | None = None):
+    logger.info("initialising model=%s model_provider=%s tools=%s", model.name, model.provider, tools)
+    chat_model = init_chat_model(model=model.name, model_provider=model.provider)
```
a-gleeson, line 28: `init_chat_model` takes `configurable_fields`  [https://python.langchain.com/api_reference/langchain/chat_models/langchain.chat_models.base.init_chat_model.html#init-chat-model](parameter). Do you think we should support passing this in? It does warn of a security risk being able to change the base_url and redirect requests. The only initial use case I can think of is over writting the base url for `ollama` if you are running the service on apple silicon and the service in a docker container, as you have to route from the docker container to the host machine.

django_app/redbox_app/redbox_core/models.py
gecBurton: > different AiSettings for different users could mean they have different models registered

I would prefer to keep this as simple as possible, would my proposal be acceptable for now and we can revisit this in the near future? 

> maintaining Embedding models in the same table

I like this idea, it is the natural next step. Please could we add this as a separate PR?

django_app/redbox_app/redbox_core/consumers.py
gecBurton: yes, you are right, i only removed it as i was too lazy to address the async issues!

redbox-core/redbox/chains/components.py
gecBurton: this sounds reasonable. how about:

```python
chat_model = init_chat_model(model=model.name, model_provider=model.provider, configurable_fields=["base_url"])
```

to begin with and we can add more if we need them?

gecBurton: [3177410](https://github.com/i-dot-ai/redbox/pull/1084/commits/317741033e571a91e24a01e2fe80bb24460d2de5)

django_app/redbox_app/redbox_core/consumers.py
gecBurton: [79de87e](https://github.com/i-dot-ai/redbox/pull/1084/commits/79de87e32060e2b01b87980d6e42e86388f7effa)

django_app/redbox_app/redbox_core/models.py
a-gleeson: That's fair and happy to go with the simpler option first and change as needed. 

Makes sense to me, I thought trying to do the embedding model here may have been too much.

django_app/redbox_app/redbox_core/models.py
```diff
@@ -51,22 +51,54 @@ def sanitise_string(string: str | None) -> str | None:
...
+        TOGETHER = "together"
+        MISTRALAI = "mistralai"
+        HUGGINGFACE = "huggingface"
+        GROQ = "groq"
+        OLLAMA = "ollama"
+
+    name = models.CharField(max_length=128, help_text="The name of the model, e.g. “gpt-4o”, “claude-3-opus-20240229”.")
+    provider = models.CharField(max_length=128, choices=Providers, help_text="The model provider")
+    description = models.TextField(null=True, blank=True, help_text="brief description of the model")
+    is_default = models.BooleanField(default=False, help_text="is this the default llm to use.")
```
a-gleeson, line 76: should we add a enabled field so we can enable and disable models as required? Then update the query to only enclude enabled models?

gecBurton: yes, good idea

gecBurton: [3c15620](https://github.com/i-dot-ai/redbox/pull/1084/commits/3c15620267e3522d3b42e40b58b2172a25883299)

django_app/redbox_app/redbox_core/views/chat_views.py
```diff
@@ -54,11 +54,11 @@ def get(self, request: HttpRequest, chat_id: uuid.UUID | None = None) -> HttpRes
...
             "llm_options": [
                 {
-                    "name": llm,
-                    "default": llm == AbstractAISettings.ChatBackend.GPT_4_OMNI.value,
+                    "name": llm.name,
+                    "default": llm.is_default,
                     "selected": llm == chat_backend,
                 }
-                for _, llm in AbstractAISettings.ChatBackend.choices
+                for llm in ChatLLMBackend.objects.all()
```
a-gleeson, line 61: filter on enabled now as it is a flag?

a-gleeson (approved):
LGTM!

django_app/redbox_app/redbox_core/views/chat_views.py
gecBurton: good point!

gecBurton: [3b62d91](https://github.com/i-dot-ai/redbox/pull/1084/commits/3b62d91ed62d14da58934385cdc85ec3b47c1add)

jamesrichards4 (approved):

PR #1087, opened 2024-10-09
Feature/redbox 697 wikipedia tool

jamesrichards4:
## Context

As an AI engineer I want Redbox to be able to use content from relevant Wikipedia pages as part of it's context when responding

Add a new Wikipedia tool which searches for relevant pages and adds a document for each page to the state documents.
Updates to the presentation of citations to allow inclusion of 'external_citations' these don't reference files in Redbox and are just URLs for another resource

## Guidance to review

Does the external citations setup work? I can add the data model stuff around that if we're happy but we should try to get this right for this and the .gov.uk tool.

The implied workflow here is that tools add documents to the state using the metadata fields as below:

metadata={
    "uuid": As normal a unique UUID for each page/element/whatever,
    "index": As normal ordered id for multiple docs in a single page/element/whatever,
    "file_name": The URL to the resource (Wikipedia page, .gov.uk page),
    "token_count": Calculated token count as normal,
    "source": "A unique name per tool added to each doc to disambiguate them and for front end labelling"
}

These can then be stored in an ExternalCitation model with fields like:

url = models.TextField()
chat_message = models.ForeignKey("ChatMessage", on_delete=models.CASCADE)
text = models.TextField()
redbox_source = models.TextField()

Details can be worked out on the PR when this is done but does this look like a good direction? Particularly to those with more Django experience, are there better ways to do this?

## Things to check


redbox-core/redbox/graph/root.py
```diff
@@ -198,10 +198,10 @@ def get_agentic_search_graph(tools: dict[str, StructuredTool], debug: bool = Fal
     )
     builder.add_edge("p_stuff_docs_agent", "p_report_sources")
     builder.add_edge("p_give_up_agent", "p_report_sources")
+    builder.add_edge("p_stuff_docs_agent", END)
+    builder.add_edge("p_give_up_agent", END)
```
rachaelcodes, line 202: Out of interest, why do we need these new endings? Is it for cases where document search takes too long?

django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -238,10 +241,24 @@ async def handle_metadata(self, response: dict):
         self.metadata = metadata_reducer(self.metadata, RequestMetadata.model_validate(response))
 
     async def handle_documents(self, response: list[Document]):
-        s3_keys = [doc.metadata["file_name"] for doc in response]
-        files = File.objects.filter(original_file__in=s3_keys)
-
+        sources = set(doc.metadata["file_name"] for doc in response)
+        files = File.objects.filter(original_file__in=sources)
```
rachaelcodes, line 245: I wonder if it's more readable to create both the file list and the non-file-source list in a single loop, e.g.:
```python
for source in sources: 
    if File.objects.filter(original_file=source).exists:
        files.add(File.object.get(original_file=source)  # Maybe there's an all in one for this line, will have a think
    else:
        external_sources.add(doc)
```

rachaelcodes:
I think the work @saisakul is doing on Citations might dramatically change the expected shape of a citation (and possibly mean we don't need to worry about all the 'put everything in a set to avoid repeats' stuff in Django), so it's probably worth a catchup there.

redbox-core/redbox/graph/nodes/tools.py
```diff
@@ -120,3 +124,43 @@ def _search_documents(query: str, state: Annotated[dict, InjectedState]) -> dict
...
+        response = _wikipedia_wrapper.load(query)
+        mapped_documents = [
+            Document(
+                page_content=doc.page_content,
+                metadata={
+                    "uuid": uuid4(),
+                    "index": i,
+                    "file_name": doc.metadata["source"],
+                    "token_count": len(tokeniser.encode(doc.page_content)),
+                    "source": "Wikipedia",
```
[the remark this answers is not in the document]
jamesrichards4: ah yes good catch, I was using ChunkMetadata which will generate a created date but the current implementation won't do that. I'll make sure it does, I think maybe additional ChunkMetadata classes would help here.

redbox-core/redbox/graph/root.py
jamesrichards4: Explicitly pointing to END is usually for readability of both the code but more importantly the visualisation. Every node sort of implicitly goes to END anyway

jamesrichards4:
> I think the work @saisakul is doing on Citations might dramatically change the expected shape of a citation (and possibly mean we don't need to worry about all the 'put everything in a set to avoid repeats' stuff in Django), so it's probably worth a catchup there.

been meaning to catchup on this anyway so I'll make sure to do that before we go any further with this

django_app/redbox_app/redbox_core/models.py
```diff
@@ -747,6 +747,20 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
         super().save(force_insert, force_update, using, update_fields)
 
 
+class ExternalCitation(UUIDPrimaryKeyBase, TimeStampedModel):
```
gecBurton, line 750: @jamesrichards4 im thinking that there are some big advantages to just extending the existing Citation model, what do you think?

```python
class Citation(UUIDPrimaryKeyBase, TimeStampedModel):
    class Origins(models.TextChoices):
        USER_FILES = "USER-FILES", _("User Files")
        WIKIPEDIA = "WIKIPEDIA", _("Wikpedia")
        GOV_UK = "GOV.UK", _("gov.uk")

    chat_message = models.ForeignKey("ChatMessage", on_delete=models.CASCADE)
    text = models.TextField()
    origin = models.TextField() # used to be called creator
    url = models.URLField(null=True, blank=True)
    file = models.ForeignKey(File, on_delete=models.CASCADE, null=True, blank=True)
    detail = JSONField(null=True, blank=True, help_text="could be page-number or uri fragment")
    origin = models.CharField(max_length=32, choices=Origins)

    def uri(self):
        return self.url or self.file.s3_url

    def save(self, ...):
        instance = super().save(...)
        if not instance.url and not instance.file:
            raise ValueError("either one of url or file must be set")
        return instance
```

KevinEtchells: I'm also thinking the same. There's enough to distinguish between internal/external citations on the frontend (should we need to), e.g. `page_numbers` vs `url`.

gecBurton: also, why dont we break this PR up into 2 bits, the redbox-core bit and the django bit?

gecBurton: @KevinEtchells im thinking that the `origin` field should be the explicit internal/external source-of-truth? (in principle a well formed `uri` should be sufficient, in practice i dont think such things really exist!)

KevinEtchells: How would this differ from the URL field?

gecBurton: uri can begin with `file://` (and other schemes such as s3://) whereas url must begin with `http(s)://`

jamesrichards4: I like the idea of a single model if we're confident it will provide the right details to break out the differences, the origin field feels like it does that to me but it's hard to be sure.

This should absolutely all have been broken down more, but now that it's here is it simpler to go ahead as is? I can break it out if you'd prefer though?

gecBurton: sry, just seen your reply James, ive addd a PR with my suggested change here https://github.com/i-dot-ai/redbox/pull/1119 i can merge into yours?

django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -255,10 +265,30 @@ async def handle_activity(self, response: dict):
...
+            handled_sources.add(file.unique_name)
+
+        additional_sources = [doc for doc in response if doc.metadata["original_resource_ref"] not in handled_sources]
+
+        for additional_source in additional_sources:
+            url = additional_source.metadata["original_resource_ref"]
+            source = additional_source.metadata.get("creator_type", "Unknown")
+            await self.send_to_client("source", {"url": url, "original_file_name": f"{source} - {url.split("/")[-1]}"})
+            self.citations.extend(
+                [(None, doc) for doc in response if doc.metadata["original_resource_ref"] == file.unique_name]
```
gecBurton, line 290: `file` isnt defined?

django_app/redbox_app/redbox_core/migrations/0053_externalcitation.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 5.1.2 on 2024-10-21 13:29
```
gecBurton, line 1: delete these, see https://github.com/i-dot-ai/redbox/pull/1122

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -21,7 +21,7 @@
 from redbox.models.graph import ROUTE_NAME_TAG, SOURCE_DOCUMENTS_TAG, RedboxActivityEvent, RedboxEventType
 from redbox.transform import combine_documents, flatten_document_state
 
-log = logging.getLogger()
+log = logging.getLogger(__name__)
```
gecBurton, line 24:
```suggestion
logger = logging.getLogger(__name__)
```

for consistency

redbox-core/redbox/models/file.py
```diff
@@ -22,12 +22,21 @@ class ChunkMetadata(BaseModel):
...
+class UploadedFileMetadata(ChunkMetadata):
+    """
+    Model for uploaded document chunk metadata.
+    """
+
+    page_number: int | None = None
     name: str
     description: str
     keywords: list[str]
+    creator_type: str = "uploaded_file"
```
gecBurton, line 42:
```suggestion
    creator_type: Literal["Wikpedia", "USER UPLOADED DOCUMENT", "gov.uk"] = "USER UPLOADED DOCUMENT"
```

to match https://github.com/i-dot-ai/redbox/blob/40f0914e25b2e6d473151e563dfe6625b0512bed/django_app/redbox_app/redbox_core/models.py#L735

django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -255,10 +265,30 @@ async def handle_activity(self, response: dict):
...
         async for file in files:
             await self.send_to_client("source", {"url": str(file.url), "original_file_name": file.original_file_name})
-        for file in files:
-            self.citations.append((file, [doc for doc in response if doc.metadata["file_name"] == file.unique_name]))
+            self.citations.append(
+                (file, sources_by_resource_ref[file.unique_name])
+            )
+            handled_sources.add(file.unique_name)
+
+        additional_sources = [doc for doc in response if doc.metadata["original_resource_ref"] not in handled_sources]
```
gecBurton, line 283: tbh i dont get this logic - what is special about the first citation that only it is treated as if it has a file?

redbox-core/redbox/test/data.py
```diff
@@ -35,9 +35,9 @@ def generate_docs(
     redbox.retriever.retrievers.hit_to_doc().
     """
     for i in range(number_of_docs):
-        core_metadata = ChunkMetadata(
+        core_metadata = UploadedFileMetadata(
             index=index_start + i,
-            file_name=s3_key,
+            original_resource_ref=s3_key,
```
gecBurton, line 40: does this mean we need to run a reingest?

redbox-core/redbox/models/file.py
```diff
@@ -22,12 +22,21 @@ class ChunkMetadata(BaseModel):
...
     uuid: UUID = Field(default_factory=uuid4)
-    index: int
-    file_name: str
-    page_number: int | None = None
+    index: int = 0  # The order of this chunk in the original resource
     created_datetime: datetime.datetime = datetime.datetime.now(datetime.UTC)
-    token_count: int
     chunk_resolution: ChunkResolution = ChunkResolution.normal
+    creator_type: str
+    original_resource_ref: str  # URL or file name
```
gecBurton, line 29:
```suggestion
    uri: FileUrl | HttpUrl  # [URL or file name](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier)
```

redbox-core/redbox/models/file.py
```diff
@@ -22,12 +22,21 @@ class ChunkMetadata(BaseModel):
...
 
     uuid: UUID = Field(default_factory=uuid4)
-    index: int
-    file_name: str
-    page_number: int | None = None
+    index: int = 0  # The order of this chunk in the original resource
     created_datetime: datetime.datetime = datetime.datetime.now(datetime.UTC)
-    token_count: int
     chunk_resolution: ChunkResolution = ChunkResolution.normal
+    creator_type: str
```
gecBurton, line 28:
```suggestion
    creator_type: Literal["..."]
```
to match the equivalent in django ?

django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -255,10 +265,30 @@ async def handle_activity(self, response: dict):
         self.activities.append(RedboxActivityEvent.model_validate(response))
 
     async def handle_documents(self, response: list[Document]):
-        s3_keys = [doc.metadata["file_name"] for doc in response]
-        files = File.objects.filter(original_file__in=s3_keys)
-
+        sources_by_resource_ref = {}
```
gecBurton, line 268:
```python
sources_by_resource_ref = defaultdict(list)
for document in response:
    ref = document.metadata.get("original_resource_ref")
    sources_by_resource_ref[ref].append(document)
```

or `groupby`?

django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -67,6 +68,7 @@ def escape_curly_brackets(text: str):
 class ChatConsumer(AsyncWebsocketConsumer):
     full_reply: ClassVar = []
     citations: ClassVar = []
+    external_citations: ClassVar = []
```
saisakul, line 71: External citations as in not the documents from user? So gov.uk would be considered external citations?

django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -255,10 +265,30 @@ async def handle_activity(self, response: dict):
...
-        files = File.objects.filter(original_file__in=s3_keys)
-
+        sources_by_resource_ref = {}
+        for document in response:
+            ref = document.metadata.get("original_resource_ref")
+            docs_from_source = sources_by_resource_ref.get(ref, [])
+            docs_from_source.append(document)
+            sources_by_resource_ref[ref] = docs_from_source
+        handled_sources = set()
+        files = File.objects.filter(original_file__in=sources_by_resource_ref.keys())
```
gecBurton, line 275: quite complicated, how about?

```python
for ref, sources in sources_by_resource_ref:
    try:
        file = File.objects.filter(original_file=ref)
        await self.send_to_client(
            "source",
            {"original_file_name": file.original_file_name},
        )
        self.citations.append((file, sources_by_resource_ref[file.unique_name]))
    except File.DoesNotExist:
        await self.send_to_client("source", {"url": ref})

        self.citations.extend(
            [(None, doc) for doc in sources]
        )
```

saisakul: Should we add type in the `sources_by_resource_ref` so it contains document's type whether it is a file/uploaded document. 

```
sources_by_resource_ref = {'ref': {'type': 'uploaded_doc', 'documents': [document]}}
```

gecBurton (approved):

PR #1089, opened 2024-10-10
Automate adding an alias to the ElasticSearch client

rachaelcodes:
## Context
Recent changes in #1071 have broken the local environment for users, as redbox-core was making reference to an Elastic alias that did not exist without running the `add_es_alias` management command. 

* Creates the chunk alias (and a default index) on the creation of the ElasticSearch client. 
* Fixes an issue with the 'delete es indices' management command that left it easy for a user to accidentally delete all chunk data from Elastic. 

## Guidance to review
Is this the right approach? Is it overkill?

## Things to check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ safe:  ##
 
 .PHONY: check-migrations
 check-migrations: stop  ## Check types in redbox and worker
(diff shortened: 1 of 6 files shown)


gecBurton (approved):
looks sensible to me

PR #1094, opened 2024-10-11
fix bug metadata only return 3 tags

saisakul:
## Context

Fixing bug in metadata: 

Invalid 'messages[1].content': string too long. Expected a string with maximum length 1048576, but got a string with length 1252301 instead

Update the metadata prompt to only return 'name', 'description' and 'keywords',

## Things to check

- [x] ~~I have added any new ENV vars in all deployed environments~~
- [x] I have run integration tests (https://github.com/i-dot-ai/redbox/actions/runs/11291983317)


gecBurton (approved):
LGTM

PR #1107, opened 2024-10-16
chore/changed model name

gecBurton:
## Context

As an Engineer I want to rename a pydantic field name so that I do not have see this error message all the time:

```
Field "model_name" in LLMCallMetadata has conflict with protected namespace "model_".

You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
```

## Things to check

### DIFF
--- a/redbox-core/redbox/models/chain.py
+++ b/redbox-core/redbox/models/chain.py
@@ -147,7 +147,7 @@ class RedboxQuery(BaseModel):
 
 class LLMCallMetadata(BaseModel):
     id: str = Field(default_factory=lambda: str(uuid4()))
-    model_name: str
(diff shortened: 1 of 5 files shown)


KevinEtchells (approved):

redbox-core/redbox/models/chain.py
```diff
@@ -147,7 +147,7 @@ class RedboxQuery(BaseModel):
 
 class LLMCallMetadata(BaseModel):
     id: str = Field(default_factory=lambda: str(uuid4()))
-    model_name: str
+    llm_model_name: str
```
gecBurton, line 150: this is the key change, all other changes are a consequence of this

PR #1111, opened 2024-10-17
Alters workflows to use OIDC auth

dannyburke1:
## Context

This change removes the need for AWS Access Keys in the workflows and moves us to an OIDC assume role model, meaning temporary access keys are used on each GHA job. We have all of the base configured in our core GHA repo so these downstream changes are minimal.

## Guidance to review

Review that the "AWS Authenication" step within the pipeline works as expected and we don't hit any snags trying to build and or deploy the image.

## Things to check


gecBurton (approved):
LGTM

PR #1113, opened 2024-10-18
Bugfix/allow no meta data

gecBurton:
## Context

As a user I do not want the file ingest to fail just because RB failed to enhance the metadata

The actual error is this:

``` 2024-10-18 11:20:35,514 ERROR monitor: Failed 'redbox_app.worker.ingest' (Bookings_v4KpDVw.xlsx) - Some error happened in metadata extraction. Error code: 400 - {'error': {'message': "Invalid 'messages[1].content': string too long. Expected a string with maximum length 1048576, but got a string with length 1239399 instead.", 'type': 'invalid_request_error', 'param': 'messages[1].content', 'code': 'string_above_max_length'}} : Traceback (most recent call last):```

So although the proposed fix doesnt address the underlying issue it will at least allow users to use their data until we find a proper solution

## Things to check

### DIFF
--- a/redbox-core/redbox/loader/loaders.py
+++ b/redbox-core/redbox/loader/loaders.py
@@ -173,10 +173,12 @@ class MetadataLoader:
         except ConnectionError as e:
             logger.warning(f"Retrying due to HTTPError {e}")
         except json.JSONDecodeError:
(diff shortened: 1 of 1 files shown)


KevinEtchells (approved):
LGTM

PR #1115, opened 2024-10-21
feature/graph-time-scales

gecBurton:
## Context

As an Admin I want to be able to see user metrics per week/day/hour not just per day

1. it is now possible to change the timescale of the graph, via a drop-down "select time scale", see picture
2. a small bug fix as been fixed so that the model name rather than integer id is displayed

image

## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/dash_apps/report_app.py
+++ b/django_app/redbox_app/redbox_core/dash_apps/report_app.py
@@ -63,7 +63,7 @@ app.layout = html.Div(
             id="breakdown",
             options=[
                 {"label": "route", "value": "route"},
-                {"label": "model", "value": "chat__chat_backend"},
+                {"label": "model", "value": "chat__chat_backend__name"},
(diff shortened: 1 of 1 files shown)


django_app/redbox_app/redbox_core/dash_apps/report_app.py
```diff
@@ -53,7 +63,7 @@
             id="breakdown",
             options=[
                 {"label": "route", "value": "route"},
-                {"label": "model", "value": "chat__chat_backend"},
+                {"label": "model", "value": "chat__chat_backend__name"},
```
gecBurton, line 66: this is a bug fix, `chat__chat_backend` used to be a string now its a foreign key to the LLM model

KevinEtchells (approved):
Looks good! Longer-term it'd be good to think about the accessibility of this page more, though I appreciate we may go in a different direction before then anyway.

PR #1116, opened 2024-10-21
Reset message-input height after sending message

KevinEtchells:
## Context

The message input box can expand to fit content. But the height doesn't reset again until you type a new message into it.

* Reset the message input height immediately after sending the message
* Fix a few type import errors

## Guidance to review

Send a multi-line message, watch it reset to original size.

## Things to check


gecBurton (approved):
LGTM

PR #1118, opened 2024-10-22
Footnote citations

saisakul:
## Context

We need citations to help users understand where the LLM use the information to answer users' queries.

The AI backend now tracks which documents were sent to the LLM and which pieces of text are cited separately. These are reported to the django app separately too (with a new callback). We're currently twisting these into shape to match the current sources frontend style but we have the data to do more.

All chunks now have a metadata 'uri' which is the link or filename to their source document. This is responsible for a large number of small changes and is part of the silly size of this PR.

The retrieval prompt is very important and agentic search will break if it isn't set correctly!

- Update retrieval response prompt so that it returns citations. The response now returns 1) markdown_answer and 2) citations.
- Add StructureOutput to handle citations. 
- Add `citations` to RedboxState

Example response:
```
{
            "answer": "Hello Kitty is a fictional character from Japan.",
            "citations": [
                {
                    "text_in_answer": "Hello Kitty is a fictional character from Japan.",
                    "sources": [
                        {
                            "source": "https://en.wikipedia.org/wiki/Hello_Kitty",
                            "document_name": "Hello Kitty",
                            "highlighted_text_in_source": "Hello Kitty (Japanese: ハロー・キティ, Hepburn: Harō Kiti),[6] also known by her real name Kitty White (キティ・ホワイト, Kiti Howaito),[5] is a fictional character created by Yuko Shimizu",
                            "page_numbers": ["1"],
                        }
                    ],
                }
            ],
        }
```
Outstanding
- We need to modify the way how we calculate input/output token as now the last LLM returns markdown_answer and citations. NOTE: We are underreporting output tokens in the current implementation
- Responses aren't streamed. This will be looked at next
- UI bug with footnote citations. [{{ no such element: redbox_app.redbox_core.models.File object['uri'] }}] This will be looked at before merging but is likely just a template change
- UI bug with search. Citations are repeated. Will be fixed before merge. Option here to use search/agentic in place of search - this might work better as @search is already familiar to some users?

## Guidance to review
Pay attention to `build_llm_chain()` function and `RequestMetadata`

## Things to check


django_app/redbox_app/redbox_core/models.py
```diff
@@ -734,7 +734,7 @@ class ChatRoleEnum(models.TextChoices):
 class Citation(UUIDPrimaryKeyBase, TimeStampedModel):
     class Origin(models.TextChoices):
         WIKIPEDIA = "Wikipedia", _("wikipedia")
-        USER_UPLOADED_DOCUMENT = "USER UPLOADED DOCUMENT", _("user uploaded document")
+        USER_UPLOADED_DOCUMENT = "UserUploadedDocument", _("user uploaded document")
```
gecBurton, line 737: are you sure you want this change? whilst i agree that this is better it will need:
1. a new migration to losen the field to just be a text field
2. a new data migration to change all existing instances of "USER UPLOADED DOCUMENT" to "UserUploadedDocument"
3. this migration to set the new enums 
4. a test to show that the data-migration is working

jamesrichards4: I did wonder about this. I think we need a consistent way of mapping the 'creatortype' field set by tools to a Citation.Origin in Django so we avoid this kind of thing happening a lot. It seems like it makes sense to do now as we haven't gone to prod yet? Maybe the best balance is to take this approach of matching the strings, make a test for that but revert this one back to what it was to avoid the migration?

gecBurton: I agree with your proposed change, but please can we do this in a separate PR because:

1. the change is big - and this PR is already substantial 
2. i would like to test this in dev, but its challenging to test anything that involves a migration in dev given that they are not reversible
?

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -280,17 +293,17 @@ def _tool(state: RedboxState) -> dict[str, Any]:
...
 
-                log.info(f"Invoking tool {tool_call['name']} with args {args}")
-
                 # Invoke the tool
                 try:
                     result_state_update = tool.invoke(args) or {}
                     tool_called_state_update = {"tool_calls": {tool_id: {"called": True, "tool": tool_call}}}
                     state_updates.append(result_state_update | tool_called_state_update)
                 except Exception as e:
+                    raise e
```
gecBurton, line 306: why catch this only to immediately raise it?

redbox-core/redbox/models/file.py
```diff
@@ -40,4 +40,4 @@ class UploadedFileMetadata(ChunkMetadata):
     name: str
     description: str
     keywords: list[str]
-    creator_type: Literal["uploaded_file"] = "uploaded_file"
+    creator_type: Literal["Wikipedia", "UserUploadedDocument"] = "UserUploadedDocument"
```
gecBurton, line 43:
```suggestion
    creator_type: Literal["UserUploadedDocument"] = "UserUploadedDocument"
```
surely?

redbox-core/redbox/graph/nodes/processes.py
jamesrichards4: ah good catch, we were debugging. Removed now

PR #1127, opened 2024-10-24
Feature/rationalise file name

gecBurton:
## Context

The file-name is especially significant and problematic as it has to be both human readable, and globally unique.

Currently we handle this via:
* adding a hash key to duplicate file names to keep the file names unique
* an additional `original_file_name` field to store the name of the file as the user uploaded it
* various helper methods, such as `unique_name`, `name`

I propose that we dont actually need a file name to be globally unique, we want it to be unique per user, so that:
* a user can overwrite a file if they upload one with the same name
* two users can upload a file with the same name and they will not conflict

to achieve this we should change the file name to user.email/file_name, this will have to convenient side effect of presenting the files in directories named after the user within s3.

* `orginal_file_name` -> now unused but kept for backwards compatibility
* `name` -> `file_name`, the original file name - for the benefit of the user

## Guidance to review

do we actually want this behaviour?

## Things to check


django_app/redbox_app/redbox_core/models.py
```diff
@@ -573,22 +585,12 @@ def delete_from_elastic(self):
         if es_client.indices.exists(index=index):
             es_client.delete_by_query(
                 index=index,
-                body={"query": {"term": {"metadata.file_name.keyword": self.unique_name}}},
+                body={"query": {"term": {"metadata.file_name.keyword": self.s3_key}}},
             )
 
-    def update_status_from_core(self, status_label):
```
gecBurton, line 579: unused

django_app/redbox_app/redbox_core/models.py
```diff
@@ -620,18 +622,21 @@ def url(self) -> URL | None:
...
 
     @property
-    def name(self) -> str:
-        # User-facing name
-        try:
-            return self.original_file_name or self.original_file.name
-        except ValueError as e:
-            logger.exception("attempt to access non-existent file %s", self.pk, exc_info=e)
+    def file_name(self) -> str:
+        if self.original_file_name:  # delete me?
```
gecBurton, line 626: for backwards compatability

django_app/redbox_app/redbox_core/models.py
```diff
@@ -536,9 +536,21 @@ def __init__(self, file):
...
+    """
+    return f"{instance.user.email}/{filename}"
+
+
 class File(UUIDPrimaryKeyBase, TimeStampedModel):
     status = models.CharField(choices=StatusEnum.choices, null=False, blank=False)
-    original_file = models.FileField(storage=settings.STORAGES["default"]["BACKEND"])
+    original_file = models.FileField(
+        storage=settings.STORAGES["default"]["BACKEND"],
+        upload_to=build_s3_key,
```
gecBurton, line 552: this `upload_to` is the key change

django_app/tests/views/test_document_views.py
```diff
@@ -64,29 +64,42 @@ def test_document_upload_status(client, alice, file_pdf_path: Path, s3_client):
 
 @pytest.mark.django_db()
 def test_upload_view_duplicate_files(alice, bob, client, file_pdf_path: Path, s3_client):
+    # delete all alice's files
```
gecBurton, line 67: this test has been changed significantly to capture the new intended behaviour

django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -307,10 +307,10 @@ async def handle_citations(self, citations: list[AICitation]):
             for s in c.sources:
                 try:
                     file = await File.objects.aget(original_file=s.source)
-                    payload = {"url": str(file.url), "original_file_name": file.original_file_name}
```
jamesrichards4, line 310: Don't we still want to use original_file_name here so we see 'stuff.pdf' in the frontend rather than 'me@cabinetoffice/stuff.pdf'?

jamesrichards4 (commented):
Just one query about displaying the file name and whether we want to display the s3 path version in the frontend

django_app/redbox_app/redbox_core/consumers.py
gecBurton: `file_name` will strip out `me@cabinetoffice/` and just return `stuff.pdf`

TBH i was hoping that we could just delete `` altogether in 30 days

jamesrichards4 (approved):

PR #1135, opened 2024-10-29
Fixed passing citations through in agentic search

jamesrichards4:
## Context

We aren't getting citation events in agentic search, we still see sources due to the document based reporting.

Doing a final parse and yield in the custom output parser so agentic search yields the full LLM response with answer and citations.

Added run_sync to Redbox. This will create the debug logging when used in local testing. Production runs should still rely on run()

## Guidance to review

There's some pretty major refactoring that needs to happen here, and improvements to testing too. If we don't move to a world where agentic search is our only route soon then we should look at these but for now I thing we should bear the pain of citations and document events overlapping in the django consumer and the missing tests

## Things to check


redbox-core/redbox/app.py
```diff
@@ -74,6 +74,12 @@ def __init__(
             debug=debug,
         )
 
+    def run_sync(self, input: RedboxState):
```
gecBurton, line 77: 👍

redbox-core/redbox/chains/parser.py
```diff
@@ -58,6 +58,7 @@ def _transform(self, input: Iterator[Union[str, BaseMessage]]) -> Iterator[Any]:
                         dispatch_custom_event(RedboxEventType.response_tokens, data=new_tokens)
                         field_length_at_last_run = len(field_content)
                         yield self.pydantic_schema_object.model_validate(parsed)
+        yield self.pydantic_schema_object.model_validate(parsed)
```
gecBurton, line 61: what happens when `parsed` is undefined?

jamesrichards4: ah good catch

gecBurton (approved):

PR #1147, opened 2024-10-30
feature/simplify metadata generation

gecBurton:
## Context

As an Engineer I want the code to be as simple as possible.

1. We extract just just first 10,000 characters rather than the first 100 tokens, this is good enough
2. We only ask the LLM for the metadata that we persist in Elastic, `name`, `keywords`, `description`, everything else is redundant
3. We use `PydanticOutputParser` to get langchain to do all the formatting 
4. We do away with merging metadata from multiple chunks, one is sufficient, and this was almost certainly the root cause of running out of tokens

## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox/actions/runs/11611785579


redbox-core/redbox/loader/loaders.py
```diff
@@ -135,77 +62,30 @@ def _chunking(self) -> Any:
...
 
-        elements = self._chunking()
-        if not elements:
-            self.metadata = self.default_metadata
-
-        else:
-            # Get first 1k tokens of processed document
-            first_n = self.get_first_n_tokens(elements, 1_000)
+        chunks = self._chunking()
+        first_thousand_words = "".join(chunk["text"] for chunk in chunks)[:1_000]
```
jamesrichards4, line 73: This is the first 1000 characters not tokens, I think we need to retain the get_first_n_tokens function to do that?

redbox-core/redbox/loader/loaders.py
```diff
@@ -135,77 +62,30 @@ def _chunking(self) -> Any:
...
-        except Exception as e:
-            raise Exception(f"Some error happened in metadata extraction. {e}")
 
+        parser = PydanticOutputParser(pydantic_object=GeneratedMetadata)
+        metadata_prompt = PromptTemplate(
+            template="".join(self.env.metadata_prompt) + "\n\n{format_instructions}\n\n{page_content}\n",
+            input_variables=["page_content"],
+            partial_variables={"format_instructions": parser.get_format_instructions()},
+        )
+        metadata_chain = metadata_prompt | self.llm | JsonOutputParser()
```
jamesrichards4, line 86: shouldn't we use the PydanticOutputParser from above here?

redbox-core/redbox/loader/loaders.py
```diff
@@ -135,77 +62,30 @@ def _chunking(self) -> Any:
...
-        else:
-            # Get first 1k tokens of processed document
-            first_n = self.get_first_n_tokens(elements, 1_000)
+        chunks = self._chunking()
+        first_thousand_words = "".join(chunk["text"] for chunk in chunks)[:1_000]
+        metadata = self.create_file_metadata(first_thousand_words)
+        return GeneratedMetadata.model_validate(metadata).model_dump()
 
-            # Get whatever metadata we can from processed document
-            doc_metadata = self.get_doc_metadata(chunks=elements, n=3, ignore=None, max_size=524288 - len(first_n))
```
jamesrichards4, line 154: we've dropped getting the metadata from unstructured, was that intentional? This metadata fed into the prompt before too which is how we could get document name for example

redbox-core/redbox/loader/loaders.py
gecBurton: what if it was 10,000 chars? either way its just a made up number

redbox-core/redbox/loader/loaders.py
gecBurton: yes, good point!

redbox-core/redbox/loader/loaders.py
gecBurton: > was that intentional

yep, because in `lazy_load` the unstructured-metadata is regenerated anyway, and the llm-generated-metadata is explicitly handled (i.e. the merging of the llm and unstructured metadata never did anything in the first place)

redbox-core/redbox/loader/loaders.py
gecBurton: [9873ba2](https://github.com/i-dot-ai/redbox/pull/1147/commits/9873ba2a0af6bbdf331e59330bae6b1c389dbf7d)

redbox-core/redbox/loader/loaders.py
```diff
@@ -135,77 +61,32 @@ def _chunking(self) -> Any:
...
-                self.metadata = self.default_metadata
-                logger.warning("LLM failed to extract metadata for this file")
-            else:
-                if all(key in res.keys() for key in self.required_keys):
-                    self.metadata = res
-                else:
-                    # missing keys
-                    self.metadata = self.default_metadata
+        chunks = self._chunking()
+        first_thousand_words = "".join(chunk["text"] for chunk in chunks)[:1_000]
```
gecBurton, line 72:
```suggestion
        first_thousand_words = "".join(chunk["text"] for chunk in chunks)[:10_000]
```

redbox-core/redbox/loader/loaders.py
```diff
@@ -83,9 +82,11 @@ def create_file_metadata(self, page_content: str) -> GeneratedMetadata:
             input_variables=["page_content"],
             partial_variables={"format_instructions": parser.get_format_instructions()},
         )
-        metadata_chain = metadata_prompt | self.llm | JsonOutputParser()
+        metadata_chain = metadata_prompt | self.llm
 
-        return metadata_chain.invoke({"page_content": page_content})
+        output = metadata_chain.invoke({"page_content": page_content})
+
+        return parser.invoke(output)
```
gecBurton, line 89: https://python.langchain.com/docs/how_to/output_parser_structured/

redbox-core/redbox/loader/loaders.py
jamesrichards4: but wasn't the combined metadata bringing the filename in from unstructured metadata? The LLM generated stuff can't find that now?

redbox-core/redbox/loader/loaders.py
```diff
@@ -135,77 +61,32 @@ def _chunking(self) -> Any:
...
-            return None
-        except Exception as e:
-            raise Exception(f"Some error happened in metadata extraction. {e}")
+        parser = PydanticOutputParser(pydantic_object=GeneratedMetadata)
+        metadata_prompt = PromptTemplate(
+            template="".join(self.env.metadata_prompt) + "\n\n{format_instructions}\n\n{page_content}\n",
+            input_variables=["page_content"],
+            partial_variables={"format_instructions": parser.get_format_instructions()},
+        )
+        metadata_chain = metadata_prompt | self.llm
```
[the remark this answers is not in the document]
gecBurton: I tried this - but it doesnt work

following this https://python.langchain.com/docs/how_to/output_parser_structured/ does the trick

redbox-core/redbox/loader/loaders.py
gecBurton: yes, but then it was ignored
it also always gets regenerated every time we use unstructured

gecBurton: i see your point, how about [6bcf9be](https://github.com/i-dot-ai/redbox/pull/1147/commits/6bcf9be62b857735702f687005ed9c0834f664d2) ?

jamesrichards4 (approved):

PR #1151, opened 2024-10-31
Feature/activity log for gadget

jamesrichards4:
## Context

We want to start emitting more events for the activity log. We're not sure what they are yet so we should create ways to emit messages which would correspond with the below draft design.
Screenshot 2024-10-31 at 14 17 08

Emit events for:
* Initial request - logging selected files
* Tools called - searching .govuk, searching your files...
* Results from tools called - reading wikipedia page, reading 7 snippets from your files...
* Response being generated - which model is being used

## Guidance to review

This is a skeleton of what we'll do once we have more information from UR. There's so messiness about where we want this logging layer to sit, separate nodes in the graph which we can run in parallel is neat but some of the context of what we need is intermediate information in a node. For example we can emit an activity log for the initial request in parallel as it's in the state, but looking at the results of a tool call has to be done in the tool node so we know where they came from.

Is this messy? Worth a discussion?

## Things to check


gecBurton (approved):
some of this is straying into UR/X/I territory @KevinEtchells ?

however i think at this stage we should be shipping something and then improving on it

PR #1161, opened 2024-11-04
Removed hosts from .env.example; updated dev docs to latest setup

jamesrichards4:
## Context

There are environment issues setting up Redbox for the first time and making it work in the various configurations

Remove hostnames from .env.example. In most cases hostnames will be pulled from a deployment level .env ie

.env.test overlaid with .env gives 'localhost' hostnames for unit tests and notebooks
.env.integration overlaid with .env gives docker network hostnames for CI and docker compose
.env alone gives no hostnames for use in vscode tasks which set the hosts manually to localhost for vscode launch configs

## Guidance to review

Does this interfere with other setups?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,7 +17,6 @@ DEV_MODE=true
 
 # === Database ===
 
-ELASTIC__HOST=elasticsearch
 ELASTIC__VERSION=8.11.0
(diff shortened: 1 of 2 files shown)


gecBurton (approved):
LGTM

PR #1163, opened 2024-11-04
Activities are collecting across messages

jamesrichards4:
## Context

Activity Events are collecting across chat messages leading to all previous events attached to the latest ai message

Reset the activities list to empty on receive

## Guidance to review

Is this the right fix for the issue? I'm definitely seeing previous activities on chat messages in preprod and this feels like we're building up all events in a websocket session?

Be good to revisit this long running consumer at some point too; for some of the interruptible/user prompting features we might want soon we need a single instance of the redbox graph (rather than per consumer).

## Things to check


KevinEtchells (approved):
This feels like the right approach to me. We only need the ones for the current message being sent to Django.

PR #1174, opened 2024-11-07
Feature/composable prompts

jamesrichards4:
## Context

Managing prompts is becoming a mess, we want to be able to store common information across prompts and to be able to reliably pass format information in without conflicting

Refactor all prompts to be just the task information, no system or format info should be provided. These are brought in from standard prompts system_info, persona_info, caller_info and format_instructions from the task if passed

NOTE: We want to do an additional piece of work here which is to redo the way we build AISettings each request to only use a set of sparse overrides from the db if created. That way we would always be using the default prompts from the settings unless we override them specifically in a deployment

## Things to check


redbox-core/redbox/chains/runnables.py
```diff
@@ -39,26 +39,32 @@
 def build_chat_prompt_from_messages_runnable(
     prompt_set: PromptSet,
     tokeniser: Encoding = None,
-    partial_variables: dict = None,
+    format_instructions: str = "",
+    additional_variables: dict = None,
```
gecBurton, line 43:
```suggestion
    additional_variables: dict|None = None,
```

redbox-core/redbox/chains/runnables.py
```diff
@@ -39,26 +39,32 @@
...
-        system_prompt_message = [("system", system_prompt)]
-        prompts_budget = len(_tokeniser.encode(system_prompt)) + len(_tokeniser.encode(question_prompt))
-        chat_history_budget = (
-            state["request"].ai_settings.context_window_size
-            - state["request"].ai_settings.llm_max_tokens
-            - prompts_budget
-        )
+        # Set the system prompt to be our composed structure
+        # We preserve the format instructions
+        system_prompt_message = f"""
```
gecBurton, line 59: 👍

gecBurton (approved):

PR #1179, opened 2024-11-11
Dropping additional steps in Gadget graph

jamesrichards4:
## Context

[at]gadget is slower than the base react implementation. One reason for this is we have an additional LLM call as we decide to answer in the agentic step, then proceed to answer generation

Swapped agentic step to only have two choices, call tools or generate answer. This removes the additional call to the LLM and simplifies the graph flow. We also remove the option to give up in the agentic step but retain it for when we're out of steps

## Guidance to review

The new search/agentic graph visualisation shows the changes best

## Things to check


gadget:
OFFTOPIC: Hi James, Could you please remove the tagging ("@gadget"). Every time someone does that I'm the one summoned :) Thank you!

jamesrichards4:
> OFFTOPIC: Hi James, Could you please remove the tagging ("@gadget"). Every time someone does that I'm the one summoned :) Thank you!

apologies! will make sure to avoid in future

gecBurton (approved):
LGTM - much cuter :+1:

PR #1185, opened 2024-11-12
Agent investigation

saisakul:
## Context

We need a way to routinely investigate how gadget works, so that we can identify bugs, unusual behaviour. 
To run the script, a structured csv files are required to run. These include: 
- prompt.csv containing User which are user id, and Prompts which are prompt query.
- documents.csv containing User which are user id, and Documents which are document file path.

EDIT:
We've updated this to use tests to run Redbox against the scenarios defined in a csv. The README details how this works. I'll update https://docs.google.com/spreadsheets/d/1NRSw5CuVAhlVH2vHSW5Ek41KxAacR6hdEWRI7lCxwmc/edit?gid=0#gid=0 to match this format, this is the central place we'll record our test cases

- add new command line script to run use cases, save LLM activities log, and extract key information.
EDIT:
- Add new ai_tests root directory containing tests for a scenarios generated from the cases csv

## Things to check


MarkDunne (approved):
haven't been able to run (thanks to vscode testing weirdness), but approving to unblock since no significant change to prod code

PR #1190, opened 2024-11-13
bugfix/allow metadata to fail

gecBurton:
## Context

As a User i want my file to be ingested even if the generation of metadata fails so that i can use my data 

## Things to check

### DIFF
--- a/redbox-core/redbox/loader/loaders.py
+++ b/redbox-core/redbox/loader/loaders.py
@@ -63,7 +63,7 @@ class MetadataLoader:
 
         return response.json() or []
 
-    def extract_metadata(self) -> dict:
+    def extract_metadata(self) -> GeneratedMetadata:
         """
         Extract metadata from first 1_000 chunks
         """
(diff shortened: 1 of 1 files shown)


redbox-core/redbox/loader/loaders.py
```diff
@@ -72,7 +72,10 @@ def extract_metadata(self) -> dict:
         original_metadata = chunks[0]["metadata"] if chunks else {}
         first_thousand_words = "".join(chunk["text"] for chunk in chunks)[:10_000]
 
-        metadata = self.create_file_metadata(first_thousand_words, original_metadata=original_metadata)
+        try:
+            metadata = self.create_file_metadata(first_thousand_words, original_metadata=original_metadata)
+        except TypeError:
+            metadata = GeneratedMetadata(name=original_metadata.get("filename"))
```
saisakul, line 78: Does original_metadata always have filename?

saisakul (approved):

redbox-core/redbox/loader/loaders.py
gecBurton: its not guaranteed but it is the most likely https://docs.unstructured.io/api-reference/api-services/document-elements#common-metadata-fields

PR #1195, opened 2024-11-14
Feature/basic api

gecBurton:
## Context

As an FE Engineer I would like a PoC API to test making REST calls to the django app

## Things to check


KevinEtchells (approved):
This works great - thanks George!

PR #1196, opened 2024-11-14
Bugfix/gov search irrelevant

saisakul:
## Context

GOV search API returns irrelevant pages resulting in incorrect documents to LLM. Example:
Using web search: [Search](https://www.gov.uk/search/all?keywords=UK+government+use+of+AI&order=relevance) 
Using API: https://www.gov.uk/api/search.json?q=UK+government+use+of+AI

Current findings:
The results from GOV search API is different from the results from search in the website i.e. [Search - GOV.UK](https://www.gov.uk/search) 
GOV search API returns pages with relevance score (es_score). However, the scores are higher for irrelevant pages. For example, a page about HMRC  has score of 0.434618 while DSIT document on AI has score of 0.08396282. 

- Add a function to recalculate similarity between query and page description. I use a simple technique i.e. `cosine_similarity` from scikit-learn, to calculate similarity.  

## Guidance to review

There are other solutions to this. For example:
- Add all pages as document, and do elastic search. However, we would potentially be adding irrelevant docs.

## Things to check

### DIFF
--- a/django_app/poetry.lock
+++ b/django_app/poetry.lock
@@ -2442,6 +2442,17 @@ files = [
 ]
 
 [[package]]
+name = "joblib"
+version = "1.4.2"
+description = "Lightweight pipelining with Python functions"
+optional = false
+python-versions = ">=3.8"
+files = [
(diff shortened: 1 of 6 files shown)


jamesrichards4:
Could you add a few test cases to test_tools for this? Be good to see at least one where we want to filter the page out and one where we don't? A real query and response would be ideal

PR #1198, opened 2024-11-14
Add expand/collapse button for activity

KevinEtchells:
## Context

So we can test a design concept with the activity log.

Add button to the right. When it's not expanded, it will only show the latest activity

image

image

It won't show if there's only one item in the log

## Guidance to review

Send a message, click the button

## Things to check


gecBurton (approved):
works for me :)

PR #1208, opened 2024-11-19
feature/added opensearch

gecBurton:
## Context

As an Engineer I would like to use opensearch so that all my stack can be managed by terraform in AWS

## Things to check


KevinEtchells (approved):
LGTM

PR #1215, opened 2024-11-20
Gov api return more docs

saisakul:
## Context

At the moment GOV search tool only return the top relevant page. However, it would be better to pull more relevant pages to use to produce answer - so that we can point to more references. 

Query first 100 relevant pages from GOV search API. Then returns top 5 relevant pages based on cosine similarity.

## Guidance to review

It would be better if instead of hard coding the no. of pages we want the tool to return, to:
- have that as tool parameters.
- use similarity threshold to select no of documents. 

## Things to check


jamesrichards4 (commented):
Let's pull both the retrieved_docs count and the returned_docs count out as parameters to the tool and set them based on a couple of new AISettings. That way we can experiment easily and they can just update their default settings to change this if they want

gecBurton (dismissed):
v good idea

saisakul:
> Let's pull both the retrieved_docs count and the returned_docs count out as parameters to the tool and set them based on a couple of new AISettings. That way we can experiment easily and they can just update their default settings to change this if they want

so rather pass them as additional parameters to the tools, I used the new AISettings from state that got passed during tool call.

redbox-core/redbox/models/chain.py
```diff
@@ -78,6 +86,10 @@ class AISettings(BaseModel):
     # this is also the azure_openai_model
     chat_backend: ChatLLMBackend = ChatLLMBackend()
 
+    # settings for tool call
+    num_retrieve_results: int = 100
```
jamesrichards4, line 90: could we give these more descriptive names? 

tool_govuk_retrieved_results
tool_govuk_returned_results

saisakul: updated

jamesrichards4 (approved):

PR #1234, opened 2024-11-25
Feat/implementation of opensearch

gecBurton:
## Things to check

- [ ] I have run integration tests https://github.com/i-dot-ai/redbox/actions/runs/12016708858


redbox-core/redbox/loader/ingester.py
```diff
@@ -22,31 +20,13 @@
 alias = env.elastic_chunk_alias
 
 
-def get_elasticsearch_store(es, es_index_name: str):
```
gecBurton, line 25: moved to settings.py

redbox-core/redbox/loader/ingester.py
```diff
@@ -22,31 +20,13 @@
...
-    return ElasticsearchStore(
-        index_name=es_index_name,
-        embedding=get_embeddings(env),
-        es_connection=es,
-        query_field="text",
-        vector_query_field=env.embedding_document_field_name,
-    )
-
-
-def get_elasticsearch_store_without_embeddings(es, es_index_name: str):
```
gecBurton, line 35: moved to settings.py

redbox-core/redbox/models/settings.py
```diff
@@ -19,7 +24,11 @@ class OpenSearchSettings(BaseModel):
 
     model_config = SettingsConfigDict(frozen=True)
 
-    collection_endpoint: str
+    host: str
```
gecBurton, line 27: @nboyse notice this change, is this ok?

nboyse: Would host be most appropriate or maybe just endpoint? As it also contains creds for DSIT and DBT, but I guess not streamlined with other configs

nboyse: I was thinking something along the lines of this, but maybe too complex for use case:
class OpenSearchSettings(BaseModel):
    """settings required for a aws/opensearch"""

    model_config = SettingsConfigDict(frozen=True)

    collection_endpoint: str = env.str("COLLECTION_ENDPOINT")
    parsed_url: AnyUrl = urlparse(collection_endpoint)

    collection_endpoint__username: Optional[str] = parsed_url.username
    collection_endpoint__password: Optional[str] = parsed_url.password
    collection_endpoint__host: Optional[str] = parsed_url.hostname
    collection_endpoint__port: Optional[str] = parsed_url.port

gecBurton: oh i see, in that case how about?

```python
class OpenSearchSettings(BaseModel):
    """settings required for a aws/opensearch"""

    model_config = SettingsConfigDict(frozen=True)
    connection_string: str

    @computed_field
    @property
    def username(self) -> int:
        return urlparse(self.collection_endpoint). username

    @computed_field
    @property
    def password(self) -> int:
        return urlparse(self.collection_endpoint). password

    @computed_field
    @property
    def password(self) -> int:
        return urlparse(self.collection_endpoint). password

    @computed_field
    @property
    def port(self) -> int:
        return urlparse(self.collection_endpoint). port

```

PR #1236, opened 2024-11-26
Make text-in-answer search case-insensitive

KevinEtchells:
## Context

Sometimes text isn't matched up to the footnote. This is because the text isn't always identical.

* Make the matching case-insensitive.

## Guidance to review

* Check footnotes still work okay.

## Things to check


saisakul (approved):
maybe run ruff

KevinEtchells:
Thanks Saisakul - this was after running Ruff!

PR #1256, opened 2024-12-13
COMMENTS - Empty PR

JLoweAI:
I created an empty branch and look to merge feature/remove-search into it to allow me to comment across the whole repo


redbox-core/redbox/models/chain.py
```diff
@@ -0,0 +1,280 @@
...
+
+    # Common Prompt Fragments
+
+    system_info_prompt: str = prompts.SYSTEM_INFO
+    persona_info_prompt: str = prompts.PERSONA_INFO
+    caller_info_prompt: str = prompts.CALLER_INFO
+
+    # Task Prompt Fragments
+
+    chat_system_prompt: str = prompts.CHAT_SYSTEM_PROMPT
```
JLoweAI, line 48: Do we just need one system prompt now?

redbox-core/redbox/models/chain.py
```diff
@@ -0,0 +1,280 @@
...
+    return RequestMetadata(
+        llm_calls=sorted(set(current.llm_calls) | set(update.llm_calls), key=lambda c: c.timestamp),
+        selected_files_total_tokens=update.selected_files_total_tokens or current.selected_files_total_tokens,
+        number_of_selected_files=update.number_of_selected_files or current.number_of_selected_files,
+    )
+
+
+class RedboxState(BaseModel):
+    request: RedboxQuery
+    documents: Annotated[DocumentState, document_reducer] = DocumentState()
```
JLoweAI, line 234: Are we still going to attatch documents to the state like this? Or have it be something that is more linked to the chat history?

redbox-core/redbox/models/chain.py
```diff
@@ -0,0 +1,280 @@
...
+    route_name: str | None = None
+    metadata: Annotated[RequestMetadata | None, metadata_reducer] = None
+    citations: list[Citation] | None = None
```
JLoweAI, line 237: Remove?

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -0,0 +1,319 @@
...
+                document_state[group][document] = None
+
+        document_state[group_uuid][document_uuid] = merged_document
+
+        return {"documents": DocumentState(groups=document_state), "metadata": request_metadata}
+
+    return _merge
+
+
+def build_stuff_pattern(
```
JLoweAI, line 134: I am guessing this and the chat pattern will be the only ones we need?

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -0,0 +1,319 @@
...
+    final_response_chain: bool = False,
+) -> Runnable[RedboxState, dict[str, Any]]:
+    """Returns a Runnable that uses state.request and state.documents to set state.messages.
+
+    If tools are supplied, can also set state.tool_calls.
+    """
+
+    @RunnableLambda
+    def _stuff(state: RedboxState) -> dict[str, Any]:
+        llm = get_chat_llm(state.request.ai_settings.chat_backend, tools=tools)
```
JLoweAI, line 148: Do we remove tools?

redbox-core/redbox/graph/nodes/processes.py
```diff
@@ -0,0 +1,319 @@
...
+        return {"documents": DocumentState(groups=document_state), "metadata": request_metadata}
+
+    return _merge
+
+
+def build_stuff_pattern(
+    prompt_set: PromptSet,
+    output_parser: Runnable = None,
+    format_instructions: str | None = None,
+    tools: list[StructuredTool] | None = None,
```
JLoweAI, line 138: Remove tools?

django_app/redbox_app/redbox_core/models.py
```diff
@@ -0,0 +1,923 @@
...
+    llm_max_tokens = models.PositiveIntegerField(null=True, blank=True)
+
+    # Prompts and LangGraph settings
+    max_document_tokens = models.PositiveIntegerField(null=True, blank=True)
+    self_route_enabled = models.BooleanField(null=True, blank=True)
+    map_max_concurrency = models.PositiveIntegerField(null=True, blank=True)
+    stuff_chunk_context_ratio = models.FloatField(null=True, blank=True)
+    recursion_limit = models.PositiveIntegerField(null=True, blank=True)
+
+    chat_system_prompt = models.TextField(null=True, blank=True)
```
JLoweAI, line 122: Am I right in thinking this is all needed because it was weirdly picking up empty strings and using them for the prompts? 

I wonder if this is something we could improve while refactoring? Especially as we are slimming down the number of prompts.

django_app/redbox_app/redbox_core/models.py
```diff
@@ -0,0 +1,923 @@
...
+    agentic_retrieval_question_prompt = models.TextField(null=True, blank=True)
+    agentic_give_up_system_prompt = models.TextField(null=True, blank=True)
+    agentic_give_up_question_prompt = models.TextField(null=True, blank=True)
+    condense_system_prompt = models.TextField(null=True, blank=True)
+    condense_question_prompt = models.TextField(null=True, blank=True)
+    chat_map_system_prompt = models.TextField(null=True, blank=True)
+    chat_map_question_prompt = models.TextField(null=True, blank=True)
+    reduce_system_prompt = models.TextField(null=True, blank=True)
+
+    # Elsticsearch RAG and boost values
```
JLoweAI, line 139: To delete?

redbox-core/redbox/chains/ingest.py
```diff
@@ -0,0 +1,48 @@
...
+    loader: UnstructuredChunkLoader,
+    s3_client: S3Client,
+    vectorstore: VectorStore,
+    env: Settings,
+) -> Runnable:
+    return (
+        document_loader(document_loader=loader, s3_client=s3_client, env=env)
+        | RunnableLambda(list)
+        | log_chunks
+        | RunnableLambda(partial(vectorstore.add_documents, create_index_if_not_exists=False))  # type: ignore[arg-type]
```
JLoweAI, line 47: What does this look like without the vector store?

redbox-core/redbox/loader/ingester.py
```diff
@@ -0,0 +1,115 @@
...
+    chunk_index_name = alias[:-8]  # removes -current
+
+    es.options(ignore_status=[400]).indices.create(index=chunk_index_name)
+    es.indices.put_alias(index=chunk_index_name, name=alias)
+
+
+def _ingest_file(file_name: str, es_index_name: str = alias):
+    logging.info("Ingesting file: %s", file_name)
+
+    es = env.elasticsearch_client()
```
JLoweAI, line 57: What does this look like without elasticsearch?

redbox-core/redbox/api/callbacks.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def on_chain_end(self, outputs: dict[str, Any], **kwargs: Any) -> None:  # noqa:ARG002
+        """Run when chain ends running."""
+        self.logger.info("Chain end: %s", outputs)
+
+    def on_chain_error(self, error: BaseException, **kwargs: Any):  # noqa:ARG002
+        """Run when chain errors."""
+        self.logger.error("Chain error: %s", error)
+
+    def on_text(self, text: str, **kwargs: Any) -> None:  # noqa:ARG002
```
JLoweAI, line 32: Is this where message logging happens?

redbox-core/redbox/models/chain.py
gecBurton: maybe two (one for chat-with-docs, the other for without?)

redbox-core/redbox/models/chain.py
gecBurton: TBC I think? 
* do we want the docs to be ordered? 
* so we want the docs to be part of the conversation so that it makes sense for a user to talk about "my last document"?
* how do we handle the case when the total doc size exceede that of the context window? 

my gut feeling is that we do want the docs to be part of the chat history but good to make this clear (if indeed it is the case)?

django_app/redbox_app/redbox_core/models.py
gecBurton: we added this to enable us to test different system prompts without having to redeploy the code. in practice it hasn't been very useful and could be removed

redbox-core/redbox/chains/ingest.py
gecBurton: TBC. the text does need to be stored somewhere, the "obvious" answer is that it should be a `TextField` on the `File` model in postgres/Django ?

redbox-core/redbox/loader/ingester.py
gecBurton: like this https://github.com/gecBurton/dosac/blob/c86f04daf96daea7b18b6ee431e9ef1e8c03ba3f/core/models.py#L101 (without the embeddings)?

redbox-core/redbox/models/chain.py
JLoweAI: Yes, good point!

redbox-core/redbox/models/chain.py
JLoweAI: I have the same gut feeling. I don't know how well the model deals with a situation that looks like: 

User: Question about a document [attaches doc]
AI: Answer about a document 
User: Question about different document [attaches doc]
AI: Answer about a document 

If the documents are only attached in the latest prompt and not in the rest of the history?

JLoweAI: But this feels like a big change that needs to be thought through. Would love @jamesrichards4 view on it

redbox-core/redbox/chains/ingest.py
JLoweAI: That makes sense to me. I thought we already saved the text? Or was it just that we saved all the chunks, and would load them all and concatenate them?

PR #1260, opened 2024-12-23
Feature/remove elastic

gecBurton:
## Things to check


redbox-core/redbox/loader/ingester.py
```diff
@@ -1,106 +1,25 @@
...
-from langchain_elasticsearch.vectorstores import BM25Strategy, ElasticsearchStore
 
-from redbox.chains.components import get_embeddings
-from redbox.chains.ingest import ingest_from_loader
 from redbox.loader.loaders import UnstructuredChunkLoader
 from redbox.models.settings import get_settings
-from redbox.models.file import ChunkResolution
+from redbox.models.file import ChunkResolution, UploadedFileMetadata
 
 if TYPE_CHECKING:
```
JLoweAI, line 9: What's this section doing? Do we need it?

JLoweAI (commented):
Thanks George, lots of nice changes here! I have left a few comments

redbox-core/redbox/models/settings.py
```diff
@@ -56,34 +47,13 @@ class ChatLLMBackend(BaseModel):
 class Settings(BaseSettings):
     """Settings for the redbox application."""
 
-    embedding_openai_api_key: str = "NotAKey"
```
JLoweAI, line 59: I imagine it's not strictly necessary, but should we remove these from the terraform code (variables and then where those variables are called)?

redbox-core/redbox/graph/root.py
```diff
@@ -241,10 +240,10 @@ def get_root_graph(
     # Subgraphs
     chat_subgraph = get_chat_graph(debug=debug)
     cwd_subgraph = get_chat_with_documents_graph(
-        all_chunks_retriever=all_chunks_retriever,
+        retriever=retriever,
         debug=debug,
     )
-    metadata_subgraph = get_retrieve_metadata_graph(metadata_retriever=metadata_retriever, debug=debug)
+    metadata_subgraph = get_retrieve_metadata_graph(retriever=retriever, debug=debug)
```
JLoweAI, line 246: What is the metadatasubgraph? Do we need it, and how does it differ from the chat with documents subgraph?

redbox-core/redbox/models/settings.py
gecBurton: good point but we are either using:
* Azure/OpenAI: which is not managed under terraform
* Bedrock: which is managed under terraform but at a team rather than a project level

redbox-core/redbox/graph/root.py
gecBurton: this is defined here https://github.com/i-dot-ai/redbox/blob/1cb033c4544372d8fc35d4135b03e647651bd73b/redbox-core/redbox/graph/root.py#L209 it is primarily used to retrieve the token "length" of the documents so that we know which strategy to use when querying docs:
* chat with docs small 
* chat with docs large (map-reduce)
* RAG (search)
* error (too large)

as we remove search and map-reduce the need for this will get simpler but we will still need to know whether the token size is too large (although we almost certainly dont need a graph to compute this)

redbox-core/redbox/loader/ingester.py
gecBurton: this is defined here https://github.com/i-dot-ai/redbox/blob/1cb033c4544372d8fc35d4135b03e647651bd73b/redbox-core/redbox/models/file.py#L39 it was used by [at]gadget to give the ReAct agent some information about the files available to it

redbox-core/redbox/models/settings.py
JLoweAI: Does that mean even more reason to remove them? Why are they currently defined as variables in terraform?

redbox-core/redbox/graph/root.py
JLoweAI: Thanks George. I am trying to work out why this is a subgraph rather than just a node. What am I missing?

redbox-core/redbox/loader/ingester.py
gecBurton: sorry - i think i missunderstood - this section `TYPE_CHECKING` is only used by `ruff` for type-checking the code.

redbox-core/redbox/models/settings.py
gecBurton: yes - good point - i hadnt spotted that - ill remove

redbox-core/redbox/graph/root.py
gecBurton: not sure tbh. this may be one for @jamesrichards4

JLoweAI: Fair. We can save it for a separate PR anyway. I will record it in the excel sheet

redbox-core/redbox/loader/ingester.py
JLoweAI: Ah makes sense, I haven't seen this done before. And would importing `mypy_boto3_s3.client import S3Client` outside of this if statement cause problems?

gecBurton: in principle its fine, in practice is just some bloat that doesnt actually do anything outside of testing. that said type checks can be pretty heavy-handed and i would be fine with just removing this

mitchfruin (approved):
Approving as discussed in Slack. James L has already checked the content.

PR #1264, opened 2025-01-03
Change "Sign in" to "Log in"

KevinEtchells:
## Context

The new designs will have "Log in" (rather than "Sign in") to make a clearer distinction from "Sign up". This is a change we can make now.

* Changed "Sign in" and "Sign out" text references to "Log in" and "Log out"
* Have done the same for any URLs
* Added a redirect for `/sign-in/`

## Guidance to review

* Check logging-in and logging-out works okay
* Check all user-facing references have been updated (I felt renaming the views was unnecessary, but I can do it you feel it would be better)
* Check using `/sign-in/` redirects properly

## Things to check

### DIFF
--- a/django_app/frontend/tests-web-components/utils.js
+++ b/django_app/frontend/tests-web-components/utils.js
@@ -3,7 +3,7 @@ const { exec } = require("child_process");
 require("dotenv").config();
 
 const signIn = async (page) => {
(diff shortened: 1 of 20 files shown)


gecBurton (approved):
LGTM

PR #1272, opened 2025-01-07
Feature/remove citations

gecBurton:
## Context

As an Engineer I want to remove all unused code

Note that this change removes all code that creates and uses citations but not the citation model itself which will be done in its own PR

## Things to check


KevinEtchells (approved):
LGTM

PR #1284, opened 2025-01-13
Removed settings, prompts and references to them

jamesrichards4:
## Context

We have a lot of unused settings and prompts from feature removals.

Delete unneeded AISettings and env Settings. Remove prompts which aren't used. Tidy up any remaining references. No impact on functionality

## Things to check


gecBurton (approved):
I think we can delete the opensearch.tf as well, but this can be a in a separate PR

PR #1286, opened 2025-01-15
feature/django counts own tokens

gecBurton:
## Context

As an Engineer I want django to be responsible for counting the tokens in every message so that this can be removed from the AI code base

1. a new `token_count` field has been added to the `ChatMessage` model
2. on `save` this field is populated
3. the migration that adds `token_count` also  back populate this field

## Things to check


redbox-core/redbox/api/format.py
```diff
@@ -18,18 +16,3 @@ def format_documents(documents: list[Document]) -> str:
         formatted.append(doc_xml)
 
     return "\n\n".join(formatted)
-
-
-def reduce_chunks_by_tokens(chunks: list[Document] | None, chunk: Document, max_tokens: int) -> list[Document]:
```
gecBurton, line 23: not used. removing this whilst im here

PR #1291, opened 2025-01-15
feature/markitdown

gecBurton:
## Context

As a PM I want to use https://github.com/microsoft/markitdown rather than https://docs.unstructured.io/ because:
* document processing will be faster
* we no longer need the chunking feature than unstructured provides
* document processing will be better because markdown is structured
* we can drop a microservice

* unstructured interface has been replace with MarkItDown
* unstructured infra hasnt actually been removed, this will be done in another PR

## Guidance to review

this is currently deployed to https://redbox-dev.ai.cabinetoffice.gov.uk/

## Things to check


JLoweAI (commented):
LGTM! 

A few things to consider removing unstructured from: 
1. `docker-compose.yml`
2. `task.json` and `launch.json` within the django app 
3. `.env.test`
4. `.env.integration`
5. Terraform (although I know that can sometimes be tricky without infra)

gecBurton:
> LGTM!
> 
> A few things to consider removing unstructured from:
> 
> 1. `docker-compose.yml`
> 2. `task.json` and `launch.json` within the django app
> 3. `.env.test`
> 4. `.env.integration`
> 5. Terraform (although I know that can sometimes be tricky without infra)

agreed to all of this, can we do it in a separate PR though please? the reason being is that terraform changes are not deployable via CI and if we need to pull this change quickly in prod then its easier to do it when the unstructured service is still there to use again

JLoweAI (approved):

PR #1306, opened 2025-01-21
feature/expose file position in queue

gecBurton:
## Context

As a user I want to know how many documents are ahead of mine in the queue so that i know how much longer i need to wait 

1. all ingest code has been moved in to the method `models.File.ingest`
2. a new method `models.File.position_in_queue` exposes how many items are ahead of the file in the queue
   * this is tested in a new `test_file_status_api_view_file`
    * and exposed to the front end via the existing api view `file_status_api_view`
3. a new nullable field `task` has been added to `models.File` to store the relevant task
4. `sanitise_string` needed to be moved into the `utils.py` so that `models` can import from `worker.py`

## Things to check


KevinEtchells (approved):
LGTM

PR #1310, opened 2025-01-22
Feature/simplify redbox core

jamesrichards4:
## Context

We don't need any of the process/runnable/graph resources any more. We've moved more logic to the Django side and there is very little if any decision making in redbox core.

Remove Langgraph from Redbox app class. Use a simple Runnable created from the existing primitives. Remove all other code and tests around it.

## Guidance to review

There are no Redbox core tests. This feels ok given there's a single runnable in there now which is just langchain components stitched together which don't assert/guarantee anything and we're catching the behaviour in consumer tests but maybe there's something we should be testing for in there?

## Things to check


gecBurton (dismissed):
all looks good but please can you delete https://github.com/i-dot-ai/redbox/blob/42eab04b7c5d205aeb4e2744b9107496eb008b2e/.github/workflows/python-app.yml#L24 so that the tests pass (by not running)

gecBurton (approved):

PR #1341, opened 2025-01-28
Ensure chat titles are unique

KevinEtchells:
## Context

We need to ensure chat titles are unique (for a user). This is to help with accessibility - if there are multiple links with the same name (but taking users to different places) this can cause confusion.

Ensure chat titles are unique  by appending " (x)" to the title (where x is the lowest unique integer)

## Guidance to review

* Create a new chat and send a message saying "Testing"
* Do the same again
* Check that one is called "Testing" and the other is called "Testing (1)"

## Things to check


gecBurton (approved):
lovely stuff!

PR #1342, opened 2025-01-29
Feature/user rate limits

jamesrichards4:
## Context

Users can break each others sessions by consuming our token quota on backend models. We want users to only be able to consume their share of these quotas

Add a user rate limiter which imposes a max tokens per minute limit for all users across all models

## Things to check


django_app/redbox_app/redbox_core/models.py
```diff
@@ -661,6 +662,11 @@ def to_langchain(self) -> AnyMessage:
             return AIMessage(content=escape_curly_brackets(self.text))
         return HumanMessage(content=escape_curly_brackets(self.text))
 
+    @classmethod
+    def get_since(cls, user: User, since: datetime) -> Sequence["ChatMessage"]:
+        """Returns all chat messages for a given user, with the most recent message after 'since'"""
+        return cls.objects.filter(chat__user=user, created_at__gt=since)
```
gecBurton, line 668:
```suggestion
cls.objects.filter(user=user).filter(created_at_gt=since).annotate(Sum("token_count"))["token_count__sum"] or 0
```

you can get the database to do the sum here?

gecBurton: but i think that there is a bigger problem here... the `ChatMessage.tocken_count` only includes what the user has typed, and not the inlcuded files, I think we need to make this change first?

gecBurton (approved):

PR #1350, opened 2025-01-31
Show detailed error message if token limit exceeded

KevinEtchells:
## Context

As a user I need to be informed if the docs I attach are above the token limit, and to understand what I can do about it.

* Frontend to get the token sizes for each file
* When a user sends the message, the token sizes are checked, and sending is blocked if above the limit
* Show a meaningful error message, showing individual file token sizes
* Update tests

## Guidance to review

* Set a low `MAX_TOKEN_SIZE` env value, and upload a large doc
* Check the error message reads okay
* Check it sends okay once you've increased `MAX_TOKEN_SIZE` again

## Relevant links

https://technologyprogramme.atlassian.net/browse/REDBOX-1012

## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/views/chat_views.py
+++ b/django_app/redbox_app/redbox_core/views/chat_views.py
@@ -72,6 +72,7 @@ class ChatsView(View):
                 }
                 for chat_llm_backend in ChatLLMBackend.objects.filter(enabled=True)
             ],
+            "max_tokens": settings.LLM_MAX_TOKENS,
         }
 
         return render(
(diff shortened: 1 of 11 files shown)


django_app/redbox_app/redbox_core/views/chat_views.py
```diff
@@ -72,6 +72,7 @@ def get(self, request: HttpRequest, chat_id: uuid.UUID) -> HttpResponse:
                 }
                 for chat_llm_backend in ChatLLMBackend.objects.filter(enabled=True)
             ],
+            "max_tokens": settings.LLM_MAX_TOKENS,
```
gecBurton, line 75: you can get the token size per model from `chat.chat_backend.context_window_size` and it is typically 128,000

KevinEtchells: Will change to that - thanks :-)

gecBurton (approved):

PR #1361, opened 2025-02-04
Feature/check file exists before ingesting it

gecBurton:
## Context

As an Engineer I want to catch the case where the `File`/`Task` relationship is broken because either:
* the `Task` is complete and has been deleted
* the `File` has been deleted by the user before the ingest process has completed

## Relevant links
https://technologyprogramme.atlassian.net/browse/REDBOX-762

## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/models.py
+++ b/django_app/redbox_app/redbox_core/models.py
@@ -619,7 +619,7 @@ class File(UUIDPrimaryKeyBase, TimeStampedModel):
         )
 
     def position_in_queue(self) -> int:
-        if not self.task:
+        if not self.task_id:
             return -1
         return OrmQ.objects.filter(lock__lt=self.task.lock).count()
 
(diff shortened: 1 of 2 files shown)


django_app/redbox_app/redbox_core/models.py
```diff
@@ -619,7 +619,7 @@ def get_ordered_by_citation_priority(cls, chat_message_id: uuid.UUID) -> Sequenc
         )
 
     def position_in_queue(self) -> int:
-        if not self.task:
+        if not self.task_id:
```
gecBurton, line 622: `.task` is the object itself, this will error if it doesnt exist, whereas `.task_id` is just the `id` which can be null

django_app/redbox_app/worker.py
```diff
@@ -14,7 +14,11 @@ def ingest(file_id: UUID) -> None:
     # These models need to be loaded at runtime otherwise they can be loaded before they exist
     from redbox_app.redbox_core.models import File
 
-    file = File.objects.get(id=file_id)
+    try:
```
gecBurton, line 17: this is the key change

jamesrichards4 (approved):

PR #1366, opened 2025-02-05
feature/progressive-enhancement-post-message

gecBurton:
## Context

As an Engineer I want a POST: REST alternative to the websocket so that a user can access the chat if they cannot use .js

a new route `chats/<uuid:chat_id>/message` has been added which can be used instead of the corresponding wss route

## Things to check


KevinEtchells (approved):
Thanks George :-)

PR #1387, opened 2025-02-14
feature/chat-metrics

gecBurton:
## Context

As a Department I want a CSV of the basic metrics so that I can see how Redbox is being adopted

I have added a management command that will allow us to create a daily metric calculation to produce a file `metrics.csv`

Out of scope - I have not added an endpoint/button that allows some users to access the file

## Things to check


redbox-core/redbox/models/settings.py
```diff
@@ -154,7 +154,7 @@ def s3_client(self):
                 region_name=self.aws_region,
             )
 
-        msg = f"unkown object_store={self.object_store}"
+        msg = f"unknown object_store={self.object_store}"
```
gecBurton, line 157: whilst im here

KevinEtchells (approved):

PR #1413, opened 2025-02-26
bugfix/department-business-unit

gecBurton:
## Context

As an Engineer I want to not delete all business units!

## Things to check


django_app/redbox_app/redbox_core/migrations/0092_remove_user_old_business_unit.py
```diff
@@ -5,7 +5,6 @@
 
 def back_populate_department_business_unit(apps, schema_editor):
     DepartmentBusinessUnit = apps.get_model("redbox_core", "DepartmentBusinessUnit")
-    DepartmentBusinessUnit.objects.all().delete()
```
gecBurton, line 8: this is the key change

PR #1427, opened 2025-03-07
feature/simplify-redbox-core

gecBurton:
## Context

As an Engineer I want to simplify our code such that it is possible to make systematic changes in the future

The key commit is https://github.com/i-dot-ai/redbox/pull/1427/commits/ca5b9233b28c5932d9f49ab1c54833247683c201 everything else is formatting and refactoring

We used to have two system prompts: `system_prompt_template` for general system behaviour and `question_prompt_template` which was a mixture of documents and the users question. The changes are:

1. we now have just one system prompt `system_prompt_template` which is general behaviour and documents. The users last question is now just the last item in the chat history

2. the function called `get_runnable` that created a langchain runnable our of our system prompts, chat history, this has now only initiates the llm with the system prompt and any documents and the messages are subsequently passes in as a argument when we call the LLM.

out of scope: there is a still a lot we could do to simplify this code like:
* the `Redbox` class is static, this could just be merged into `RedboxState`
* `redbox.chains.componts` could be moved into django `Settings` or even just removed when we have litellm
* `redbox.models` could be squashed into a single file

I have not done this to keep code diff focussed

## Things to check

### DIFF
--- a/django_app/redbox_app/redbox_core/consumers.py
+++ b/django_app/redbox_app/redbox_core/consumers.py
@@ -23,7 +23,7 @@ logger.info("WEBSOCKET_SCHEME is: %s", settings.WEBSOCKET_SCHEME)
 
 
 class ChatConsumer(AsyncWebsocketConsumer):
-    redbox = Redbox(debug=settings.DEBUG)
+    redbox = Redbox()
 
     async def receive(self, text_data=None, _bytes_data=None):
         """Receive & respond to message from browser websocket."""
--- a/redbox-core/redbox/app.py
+++ b/redbox-core/redbox/app.py
@@ -1,6 +1,6 @@
 from logging import getLogger
 
-from langchain_core.messages import AIMessage
+from langchain_core.messages import AIMessage, BaseMessage
 
 from redbox.models.chain import RedboxState
 
@@ -13,7 +13,7 @@ logger = getLogger(__name__)
 
 
 class Redbox:
-    def run_sync(self, state: RedboxState):
+    def run_sync(self, state: RedboxState) -> BaseMessage:
         """
         Run Redbox without streaming events. This simpler, synchronous execution enables use of the graph debug logging
         """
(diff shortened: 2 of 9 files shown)


django_app/redbox_app/redbox_core/consumers.py
```diff
@@ -23,7 +23,7 @@
 
 
 class ChatConsumer(AsyncWebsocketConsumer):
-    redbox = Redbox(debug=settings.DEBUG)
```
gecBurton, line 26: we dont actually use this variable anymore

redbox-core/redbox/app.py
```diff
@@ -15,51 +13,20 @@ async def _default_callback(*args, **kwargs):
...
-        request_dict = state.model_dump()
-        return self._get_runnable(state).invoke(input=request_dict)
+        return state.get_llm().invoke(input=state.get_messages())
 
     async def run(
         self,
         state: RedboxState,
         response_tokens_callback=_default_callback,
-    ) -> RedboxState:
+    ) -> AIMessage:
```
gecBurton, line 26: this was a bug

redbox-core/redbox/app.py
```diff
@@ -15,57 +13,25 @@ async def _default_callback(*args, **kwargs):
...
-        messages = (
-            [settings.system_prompt_template]
-            + state.messages[:-1]
-            + PromptTemplate.from_template(settings.question_prompt_template, template_format="jinja2")
-            .invoke(input=input_state)
-            .to_messages()
-        )
-        return ChatPromptTemplate.from_messages(messages=messages) | llm
-
     def run_sync(self, state: RedboxState):
```
JLoweAI, line 16: Does this need a `-> AIMessage`?

JLoweAI (approved):
LGTM!

redbox-core/redbox/app.py
gecBurton: good idea [cefae37](https://github.com/i-dot-ai/redbox/pull/1427/commits/cefae372f71f774c93a209d18c70eb1c599e3706)

